PR #99, opened 2020-12-07
SOQL code completion for SELECT fields

dobladez:
### What does this PR do?

Provide initial code completion with fields on SELECT clause

W-8432064

### DIFF
--- a/packages/language-server/src/completion.ts
+++ b/packages/language-server/src/completion.ts
@@ -1,32 +1,50 @@
-import { SoqlParser } from '@salesforce/soql-parser/lib/generated/SoqlParser';
+import {
+  SoqlFromExprsContext,
+  SoqlParser,
+  SoqlQueryContext,
+} from '@salesforce/soql-parser/lib/generated/SoqlParser';
 import { SoqlLexer } from '@salesforce/soql-parser/lib/generated/SoqlLexer';
+import { SoqlParserListener } from '@salesforce/soql-parser/lib/generated/SoqlParserListener';
 import { LowerCasingCharStream } from '@salesforce/soql-parser';
-import { CompletionItemKind } from 'vscode-languageserver';
+import {
+  CompletionItem,
+  CompletionItemKind,
+  InsertTextFormat,
+} from 'vscode-languageserver';
 
-import { CommonTokenStream, TokenStream } from 'antlr4ts';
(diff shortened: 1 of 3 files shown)


packages/language-server/src/completion.ts
```diff
@@ -149,3 +248,96 @@ function isCursorAfter(
...
+  }
+  private foundMatch(fromToken: Token, sobjectName: string) {
+    const selectToken = this.selectsStack.pop();
+    if (selectToken) {
+      this.matchedSelectFroms.push({
+        select: selectToken,
+        from: fromToken,
+        sobjectName: sobjectName,
+      });
+    }
```
jgellin-sf, line 319: Does `SELECT (SELECT |) FROM X` break the stack-based matching? If so, won't `SELECT |, (SELECT ) FROM X` not match the first SELECT with the X object? Do we need a check to make sure the SELECT and FROM tokens have the same immediate inner query ancestor?

jgellin-sf (commented):
Solid work here. One question about the matching algorithm.

packages/language-server/src/completion.ts
dobladez: Good catch!.  Hmm... the second case you mention seems to work fine.

On the first case, it proposes completions for `X` (which is wrong) but it's not  "matching" (that is "pairing") of FROMs with SELECTs that's broken, the problem is on method `findSObjectName()` which is only looking at the matched pairs, ignoring the SELECTs remaining on the stack.
I'll try to fix it.

BTW, How about I rename `matchedSelectFroms` to `pairedSELECT_FROMs` or similar?

jgellin-sf: I'm ambivalent on the rename. Whatever works. :)

dobladez: No, I was wrong! You are right that the matching fails on the first case!
I've replied too quickly assuming the "pairing" of the FROMs was done on `exitSoqlInnerQuery` (instead of `exitSoqlFromExprs`)... so yes, the FROM is paired with the inner SELECT on your first example.

The second example does work, but for the wrong reason ;-): The second `SELECT` is not recognized by the parser due to the syntax error, thus the FROM is paired with the first SELECT.

Anyway, I'll work on a fix. Thanks again!

jgellin-sf (approved):
QA passed. The only issues I found were those already identified and to be covered by W-8558994.

PR #100, opened 2020-12-07
[WIP] Send basic telemetry if soql error or unsupported syntax

dehru:
### What does this PR do?
This basically looks for errors or unsupported syntax and sends a count to the custom editor.  It will help us track how often users are seeing "Syntax Error" and "Unsupported Syntax" in the UI.  Right now, it only sends the counts of errors / unsupported.  

This pairs with this PR in salesforcedx-vscode: https://github.com/forcedotcom/salesforcedx-vscode/pull/2795

@W-8086238@


praksb:
I am not sure how useful it would be to have just a count of unsupported syntax and errors. Also, wouldn't they always be 1.
I think it would be better to grab the query in case of unsupported syntax and send it instead so that we could figure out which syntax we need to start supporting in the builder.

For the error case, I wonder if it would be too much noise to log these and what value it will give us.

jgellin-sf:
I agree with @praksb that the actual query is more useful than the metadata this code is currently sending, but aren't there GDPR concerns with users' actual query code? A middle way is maybe to classify an error or unsupported syntax instance if we can ("unsupported:groupby", "unsupported:innerquery", etc.). That would require some model analysis and would probably be in a separate PR. 

Regardless of the content of what telemetry gets sent, the framework for telemetry message passing in this PR looks solid.

dehru:
Yes, sending more of the query was quite useful.  The original implementation i did was to send the array of errors and unsupported syntax that the parser gave me.  But then I noticed that it included the full WHERE statement like `WHERE Id = '12345'`.   So i backed it out to the much less useful counts because of GDPR.  

@jgellin-sf  I wondered if there's a way for the parser to help remove the GDPR protected bits or not.

I think It would be nice to allow the salesforcedx-vscode-soql PR to accept either a count or an array here and send that to telemetry.  And then maybe we can figure out how to send more query information once we strip out the GDPR protected bits.

jgellin-sf:
@dehru I think we could easily accommodate a sort of classification of unsupported or error syntax in the model. I've created W-8547957 as a placeholder for this work.

dehru:
Here's an example of where I discovered sending GDPR protected stuff into telemetry during my testing.

Screen Shot 2020-12-08 at 11 11 25 AM

dehru:
Combining mine and jon's PRs into one since they need to be integrated anyway.

Issue #71, opened 2020-11-11
From picklist overflows to SOQL Query preview box


lcampos:
After loading the list of SObject from the org, the From picklist gets refreshed and that causes the picklist to overflow to the SOQL Query preview box.

<img width="1144" alt="misaligned-from-picklist" src="https://user-images.githubusercontent.com/1041842/98846456-e32fe200-2403-11eb-8d0e-696e6a7d846b.png">

OS and version: Mac OS 10.15.7
VS Code Version: 1.51.0

jonnyhork:
I think we need to add a width to the <select> https://stackoverflow.com/questions/10672586/how-to-make-select-elements-shrink-to-max-width-percent-style-within-fieldset

PR #101, opened 2020-12-09
Jhork: Support for Simple WHERE Expressions

jonnyhork:
### What does this PR do?
This PR will add preliminary support for simple WHERE expressions in the SOQL Builder.

Where UI first PR

Where UI Map

**Included In this PR:**
- WHERE Expression UI Model ---> Parser Model ---> Query String ---> UI Model integration.
- Allow user to create one or more simple `fieldCompareExpressions`: `WHERE Name = 'PWT'`
- Allow users to select either `AND` or `OR` to join multiple expressions.
- Allow user to choose `= ≠ < ≤ > ≥ like` operators.
- Basic validation to wrap compare values in quotes.

Testing is still in progress.

**Future Work:**

- Error Handling - [W-8555673](https://gus.lightning.force.com/a07B0000008l9URIAY)
- Better Validation - [W-8555377](https://gus.lightning.force.com/a07B0000008l98pIAA)
- Better Support for LIKE - [W-8555421](https://gus.lightning.force.com/a07B0000008l9BoIAI)
- Prevent User Input from deleting while typing - [W-8555599](https://gus.lightning.force.com/a07B0000008l9JUIAY)  PR: https://github.com/forcedotcom/soql-tooling/pull/105

@ W-7987155@ @ W-7987137@


packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
```diff
@@ -37,6 +37,34 @@ export function convertSoqlModelToUiModel(
...
+    where.conditions = where.conditions
+      .filter((condition) => !SoqlModelUtils.containsUnmodeledSyntax(condition))
+      .map((expression, index) => {
+        const operator =
+          expression instanceof Impl.LikeConditionImpl
+            ? 'LIKE'
+            : SoqlModelUtils.getKeyByValue(
+                Soql.CompareOperator,
+                expression.operator
+              );
```
jonnyhork, line 54: This converts raw operator values to string enum values to be handled by the UI.
`=` ---> `EQ`,
`>` ----> `GT`
This will allow us to better handle translating a simple `LIKE` operator into `starts with` as a display value when the time comes.

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
```diff
@@ -37,6 +37,34 @@ export function convertSoqlModelToUiModel(
...
+        return {
+          field: expression.field.fieldName,
+          operator: operator,
+          criteria: expression.compareValue,
+          index
+        };
+      });
+
+    const andOrFromSoqlModel = queryModel.where.condition.andOr;
+    where.andOr = andOrFromSoqlModel;
```
jonnyhork, line 65: If `andOrFromSoqlModel ` is undefined then the UI will use the default `AND` value as an initial state.

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
```diff
@@ -92,6 +122,41 @@ function convertUiModelToSoqlModel(uiModel: ToolingModelJson): Soql.Query {
...
+  let whereExprsImpl;
+  if (uiModel.where && uiModel.where.conditions.length) {
+    const whereExprsArray = uiModel.where.conditions.map((where) => {
+      const criteriaImpl =
+        where.criteria instanceof Impl.LiteralImpl
+          ? where.criteria
+          : new Impl.LiteralImpl(
+              where.criteria.type,
+              `${where.criteria.value}`
+            );
```
jonnyhork, line 135: I am passing the whole `LiteralImp` object as the value because it contains useful info such as the data type we may want in the future:
`LiteralImp {type: String, value: 'PWT'}`
When there is a new condition expressions the `criteria.value` is a simple string however existing expressions have a criteria property that is already an object. So this logic is to handle that case.
Screen Shot 2020-12-01 at 2 12 57 PM

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
```diff
@@ -92,6 +122,41 @@ function convertUiModelToSoqlModel(uiModel: ToolingModelJson): Soql.Query {
...
+      return new Impl.FieldCompareConditionImpl(
+        new Impl.FieldRefImpl(where.field),
+        Soql.CompareOperator[where.operator],
+        criteriaImpl
+      );
+    });
+
+    const andOr = uiModel.where.andOr;
+    whereExprsImpl = SoqlModelUtils.arrayToSimpleGroup(whereExprsArray, andOr);
+  }
```
jonnyhork, line 153: The assumption is that the UI Modele will always have context for `AndOr` based on the UI state.

packages/soql-builder-ui/src/modules/querybuilder/services/toolingModelService.ts
```diff
@@ -142,19 +140,78 @@ export class ToolingModelService {
...
+      updatedWhereField
+    );
+    /*
+    The UI model should always be aware
+    of andOr UI state when expr is updated.
+    */
+    newModel = newModel.setIn(
+      [ModelProps.WHERE, ModelProps.WHERE_AND_OR],
+      andOr
+    );
```
jonnyhork, line 189: This is here because when the user goes from 1 condition to 2 conditions, the UI model needs to provide that context in order to convert the UIModel to ParserModel without errors. This method is called when there is a change in user input, such as typing. I did not want to send a second event or update the model twice each time to include the AndOr state, so I stuck it in here.

packages/soql-builder-ui/src/modules/querybuilder/services/toolingModelService.ts
```diff
@@ -142,19 +140,78 @@ export class ToolingModelService {
...
 
+  public setAndOr(andOr: string) {
+    const currentModel = this.getModel();
+    const newModel = currentModel.setIn(
+      [ModelProps.WHERE, ModelProps.WHERE_AND_OR],
+      andOr
+    );
+
+    this.changeModel(newModel);
+  }
```
jonnyhork, line 158: This is called when the user toggles the AND|OR buttons and:
- there are more than one expression
- the last expression is filled out

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (where.conditions && where.conditions.length) {
+      this._conditionsStore = where.conditions;
+    } else {
+      this._conditionsStore = [this.conditionTemplate];
+    }
+
+    if (where.andOr) {
+      this._andOr = where.andOr;
+    }
+  }
```
jonnyhork, line 51: `whereExpr` represents the conditions coming in from the UIModel.

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
```diff
@@ -0,0 +1,144 @@
...
+
+interface ModifierGroupNode extends Node {
+  allModifiersHaveValue: boolean;
+}
+
+export default class Where extends LightningElement {
+  @api isLoading = false;
+  @api whereFields: string[];
+  @track _conditionsStore: JsonMap[] = [];
+  _andOr = AndOr.AND;
```
jonnyhork, line 28: Represents the UI state

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+interface ModifierGroupNode extends Node {
+  allModifiersHaveValue: boolean;
+}
+
+export default class Where extends LightningElement {
+  @api isLoading = false;
+  @api whereFields: string[];
+  @track _conditionsStore: JsonMap[] = [];
```
jonnyhork, line 27: This is the property that is used to render the list of modifier groups. We use a private property so that we can add a group to the UI without changing the model until all the values are filled out.

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    if (this.getModfierGroupsRendered().length) {
+      const lastGroupIndex = modfierGroupsRendered.length - 1;
+      const lastGroupIsComplete = (modfierGroupsRendered[
+        lastGroupIndex
+      ] as ModifierGroupNode).allModifiersHaveValue;
+
+      this.lastModifierGroupIsComplete = lastGroupIsComplete;
+    }
+  }
```
jonnyhork, line 102: If the last modifier group is not complete then we do not want the user to add more groups or change the `AndOr` selection because it will remove the empty modifier group they just added.

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
```diff
@@ -0,0 +1,144 @@
...
+        this.getModfierGroupsRendered().length > 1 &&
+        this.lastModifierGroupIsComplete
+      ) {
+        const andOrSelectionEvent = new CustomEvent('andorselection', {
+          detail: selectedValue
+        });
+        this.dispatchEvent(andOrSelectionEvent);
+      }
+    }
+  }
```
jonnyhork, line 132: This allows the UI state to  change independently from changing the model when its appropriate. Such as when there is only one or no selected conditions.
I also wanted to check somewhere that the values sent to the model are only `AND` or `OR`.

And event will only be sent is there
- is more than one condition
- they are all filled out.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.css
```diff
@@ -0,0 +1,44 @@
...
+}
+/*
+setting a width will
+prevent select el from
+expanding to fit content
+ */
+.modifier__fields {
+  width: 100%;
+  flex-grow: 2;
+}
```
jonnyhork, line 30: This this will prevent the select element width from changing based on the content.

https://stackoverflow.com/questions/10672586/how-to-make-select-elements-shrink-to-max-width-percent-style-within-fieldset

Adding a width to the FROM select component might fix this issue too https://github.com/forcedotcom/soql-tooling/issues/71

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -0,0 +1,192 @@
...
+    {
+      value: 'LIKE_CONTAINS',
+      displayValue: 'contains'
+    } */
+  ];
+  handleSelectionEvent: () => void;
+  // this need to be public so parent can read value
+  @api get allModifiersHaveValue() {
+    return this._allModifiersHaveValue;
+  }
```
jonnyhork, line 71: This is used so that the `where` component can control the `add` button and other events based on the state of each child component.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  constructor() {
+    super();
+    this.handleSelectionEvent = debounce(selectionEventHandler.bind(this), 500);
+  }
```
jonnyhork, line 76: Debounce `keyup` events when the user is typing. Shorter time = snappier UI but more glitches. See [W-8555599](https://gus.lightning.force.com/a07B0000008l9JUIAY)

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -0,0 +1,192 @@
...
+    return this._criteria;
+  }
+  set criteria(criteria) {
+    if (criteria && criteria.value) {
+      const cleanedValue = criteria.value.replace(/['"]+/g, '');
+      this._criteria = { ...criteria, value: cleanedValue };
+    } else {
+      this._criteria = criteria;
+    }
+  }
```
jonnyhork, line 136: This will prevent quotes from being displayed in the text inputs.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -0,0 +1,192 @@
...
+    const conditionRemovedEvent = new CustomEvent('where__conditionremoved', {
+      detail: {
+        index: this.index
+      },
+      bubbles: true,
+      composed: true
+    });
+
+    this.dispatchEvent(conditionRemovedEvent);
+  }
```
jonnyhork, line 149: This event only needs to be handled two levels up by App.ts, so the event is set to propagate through the DOM( bubbles) and cross the shadow boundary (composed).
The event is namespaced to prevent event collisions because it propagates all the way to the document root.
https://lwc.dev/guide/events#configure-event-propagation

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -0,0 +1,192 @@
...
+  For now, this will wrap the user input in quotes unless
+  the value is:
+  - a number
+  - a boolean value
+  */
+  normalizeInput(value: string): string {
+    // prevent values from being wrapped in quotes twice
+    value = value.replace(/['"]+/g, '');
+    if (
+      !isNaN(+value) ||
```
jonnyhork, line 162: Asking, "can the value be parsed to a number"

jonnyhork: I think I will assign a variable name to this so its easier to understand.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -158,8 +158,9 @@ export default class WhereModifierGroup extends LightningElement {
   normalizeInput(value: string): string {
     // prevent values from being wrapped in quotes twice
     value = value.replace(/['"]+/g, '');
+    const canValueBeParsedToNumber = !isNaN(+value);
     if (
-      !isNaN(+value) ||
+      canValueBeParsedToNumber ||
       value.toLocaleLowerCase() === 'true' ||
       value.toLocaleLowerCase() === 'false'
     ) {
```
jgellin-sf: Maybe this is good enough for now, but this is dangerous territory. NULL and the various date formats are additional values that can be expressed without quote. ID's are generally strings (so quotes are needed) that consist of numbers.

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
```diff
@@ -37,6 +37,34 @@ export function convertSoqlModelToUiModel(
       : undefined;
 
   const sObject = queryModel.from && queryModel.from.sobjectName;
+
+  let where;
+  if (queryModel.where && queryModel.where.condition) {
+    const conditionsObj = queryModel.where.condition;
+    where = SoqlModelUtils.simpleGroupToArray(conditionsObj);
```
jgellin-sf, line 44: `SoqlModelUtils.simpleGroupToArray` will throw if the `conditionsObj` is not a simple group (simple group = does not mix AND/OR). If this is not the desired behavior, you can check first with `SoqlModelUtils.isSimpleGroup()`.

jgellin-sf (commented):
Nicely structured. 

I have some concerns about the dueling models, each getting quite a bit more complex with the support of WHERE clauses. That is maybe a matter for tech debt.

A more immediate concern is the treatment of literal values. It's too hard to make assumptions about type just from the value of the literal. We should move towards using the field metadata at our disposal ASAP.

packages/soql-builder-ui/src/modules/querybuilder/app/app.html
```diff
@@ -49,6 +57,15 @@ <h3>Syntax Error</h3>
           is-loading={isFieldsLoading}
           class={theme}
         ></querybuilder-fields>
+        <querybuilder-where
+          where-expr={query.where}
+          where-fields={fields}
+          onmodifierselectionwithandor={handleWhereSelection}
+          onandorselection={handleAndOrSelection}
+          onwhere__conditionremoved={handleRemoveWhereCondition}
```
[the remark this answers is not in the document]
dehru: `onmodifierselectionwithandor`.  Could we just call this `onwhereadded` or `onfilteradded`?  A simpler name for that event would aid high-level understanding.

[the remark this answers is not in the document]
dehru: I am really unconcerned with event bubbling and namespacing in our application.  We own the whole dom.  This will not be embedded inside another application in the same dom.  But having said that, i find the event handler for the namespace super easy to read.  it makes me think we may want to consider using underscores elsewhere.  `onwhere_removed` `onwhere_added` `onandor_changed`.  Running all the words together with lower case is just plain hard to read.

packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
```diff
@@ -169,26 +169,35 @@ export default class App extends LightningElement {
...
-
   handleOrderByRemoved(e) {
     this.modelService.removeOrderByField(e.detail.field);
   }
-
+  /* ---- LIMIT HANDLERS ---- */
   handleLimitChanged(e) {
     this.modelService.changeLimit(e.detail.limit);
   }
+  /* ---- WHERE HANDLERS ---- */
```
dehru, line 191: I think we should model the events after these handler names.

packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
```diff
@@ -40,15 +55,17 @@
 
 .selected-field {
   margin: 0.5rem 0.5rem 0 0;
-  display: inline-block;
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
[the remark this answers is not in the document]
dehru: fancy

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
dehru: This was probably useful for debugging.  But just collapse into 1 line for now.  `where.andOr = queryModel.where.condition.andOr;`

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
dehru: All these 😈 Imps 😈 everywhere!

packages/soql-builder-ui/src/modules/querybuilder/services/toolingModelService.ts
dehru: Nice work with the awkward syntax of immutablejs.  And adding ModelProps.  Agile continuous refinement.

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export default class Where extends LightningElement {
+  @api isLoading = false;
+  @api whereFields: string[];
+  @track _conditionsStore: JsonMap[] = [];
+  _andOr = AndOr.AND;
+  conditionTemplate: ConditionTemplate = {
+    field: undefined,
+    operator: undefined,
+    criteria: { type: null, value: null },
```
dehru, line 32: let's use undefined instead of null, for consistency with the field/operator undefined in the template.

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
```diff
@@ -0,0 +1,144 @@
...
+      this._conditionsStore = [this.conditionTemplate];
+    }
+
+    if (where.andOr) {
+      this._andOr = where.andOr;
+    }
+  }
+
+  get templateIndex(): number {
+    if (this._conditionsStore) {
```
dehru, line 54: Is _conditionsStore ever undefined?

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
dehru: Nice use of interface to define an additional property on node.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.html
```diff
@@ -0,0 +1,57 @@
...
+    >
+      <template if:true={hasSelectedField}>
+        <option value={selectedField}>{selectedField}</option>
+      </template>
+      <template if:false={hasSelectedField}>
+        <option value="" class="placeholder" data-el-default-option>
+          {defaultFieldOptionText}
+        </option>
+      </template>
+      <template for:each={filteredFields} for:item="field">
```
dehru, line 18: It's unfortunate that there is no logic in LWC template syntax.  Dealing with select components in our UI is overly complicated by that.

dehru (approved):
Awesome work.  It was exciting to see this evolve.  And each roadblock along the way get CRUSHED by your creative solutions!

packages/soql-builder-ui/src/modules/querybuilder/app/app.html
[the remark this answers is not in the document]
jonnyhork: yeah, I agree with everything in your second comment. The namespacing does make it easier to to parse out what the context of the event is. I'm down to give that treatment to the rest of our events.

I can think about a different name for `onmodifierselectionwithandor`, it's not just an "added" event though. It fires each time someone changes a selection or value in an existing condition as well as adding a new condition with completed info. 

I would like to be as specific as I can with the events, especially with `where` because once we have support for more nested conditions and complex queries I thought it might be nice to distinguish parts of the model/components that changed. But maybe that's not going to be the case and the where expression, no matter how complex, will packaged and sent as whole.

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
jonnyhork: I used `null` because `undefined` shows up in the `<input>` as a value, where as `null` is not rendered

dehru: oh that's interesting.

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
jonnyhork: Thats a good call, I dont think so, this is left over from earlier iterations. I don't think I need this any more.

packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
jonnyhork: So you thinking something like:
```javascript
 handleWhereSelection(e) {
    // onwhere_selection
  }
  handleAndOrSelection(e) {
    // onwhere_andor_selection
  }
  handleRemoveWhereCondition(e) {
    // onwhere_remove_condition
  }
```

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
jonnyhork: yeah I was surprised to see that too, maybe I can google it or leave it, idk.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -0,0 +1,193 @@
...
+    super();
+    this.handleSelectionEvent = debounce(selectionEventHandler.bind(this), 500);
+  }
+
+  renderedCallback() {
+    this.checkAllModifiersHaveValues();
+  }
+
+  checkAllModifiersHaveValues(): Boolean {
+    const fieldEl = this.template.querySelector('[data-el-where-field]');
```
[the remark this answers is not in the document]
jonnyhork: ok, I thought about that but got bit before with the cached values not representing the current state I wanted. It would be nice to only query for them once though. I'll try and refactor.

dehru:
> I have some concerns about the dueling models, each getting quite a bit more complex with the support of WHERE clauses. That is maybe a matter for tech debt.

@jgellin-sf - The models aren't too far apart and I added a story for the backlog to attempt to remove the separate UI model.  With the WHERE hierarchical structure of course being the most challenging.  I've done similar hierarchical ui components in the past and would be a fun challenge.

packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
```diff
@@ -25,6 +31,15 @@
...
 
+/* TODO: we should change the name of the var
+the width of all the children should be
+default to the width of the parent.
+*/
+.section_selection-container {
+  min-width: var(--soql-input-width);
+  max-width: var(--soql-input-width);
+}
+
```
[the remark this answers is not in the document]
dehru: Do we do tech debt weeks? 😆

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
jonnyhork: Could be worth discussing for sure, @jgellin-sf also brought up concerns about the dueling models.

In this case I don't really see it as issue so much. I may be wrong but the console is displaying the type of the object as `LiteralImpl` here, but isn't it still bust a simple json object:

```javascript
{
  field: "id",
  operator: "EQ",
  criteria: {
   type: '',
   value: '',
 },
 index: 0
}
```
I like this ☝️ structure because it organized the info for me, even if its a little nested.

packages/soql-builder-ui/src/modules/querybuilder/services/toolingModelService.ts
jonnyhork: Thanks!

packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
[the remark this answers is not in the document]
jonnyhork: haha doesn't seem like it anymore.

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
jonnyhork: fixed in the `simplify indexing of new conditions` commit

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
dehru: Yeah, my only complaint is seeing `LiteralImpl` in the console when i wanna see the details usually.  Really, nothing else.

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
jonnyhork: Fixed in the `catch complex condition groups as unsupported` commit

packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
jonnyhork: fixed in the `update names of all events to be more readable` and `update where event names`

PR #103, opened 2020-12-14
Better telemetry through metadata

jgellin-sf:
### What does this PR do?
This PR adds some metadata to the model that can be useful for telemetry. Currently, telemetry reports the number of unmodeled syntax objects and the number of errors in a query model. A more qualitative approach would allow us to make better inferences from the telemetric data. For example, it is more useful to know that a query has unmodeled syntax because of the presences of a semi-join than simply that it has 1 instance of unmodeled syntax.

This PR classifies unmodeled syntax objects with a `reason` property that accounts for _why_ that particular model object is unmodeled. It also adds a `grammarRule` property to `ModelError` which shows which grammar rule the error takes place in. Combined with the type of error ('UNKNOWN', 'INCOMPLETEFROM', etc), this tells us much more about an error than simply that there is one.

The changes in telemetry that these changes to the model can support are illustrated in the following examples.

Quantitative telemetry (current):
```
unsupported: 2,
errors: 2
```
Qualitative telemetry:
```
unsupported: ['unmodeled:alias', 'unmodeled:semi-join'],
errors: ['NOFROM:SoqlFromClauseContext', 'UNKNOWN:SoqlSelectExprsContext']
```

@W-8547957@


dehru:
I like the reason addition.  I'm wondering if we could also keep the descriptive test do possibly display to the user?  and when i'm sending to telemetry i will only send the reason field.
  
Screen Shot 2020-12-07 at 1 26 27 PM

jgellin-sf:
@dehru The existing properties remain, so you can still get the actual syntax that is causing the syntax to be out of support. Is this what you mean, or do you think there should be additional descriptive text?

dehru:
> @dehru The existing properties remain, so you can still get the actual syntax that is causing the syntax to be out of support. Is this what you mean, or do you think there should be additional descriptive text?

Yes, that's what i mean.  Perfect.  Thanks JG!

dehru (approved):

dehru:
Combining mine and jon's PRs into one since they need to be integrated anyway.

PR #104, opened 2020-12-15
Fix LSP dependency

jgellin-sf:
### What does this PR do?
VSCode has made some dodgy version changes with their latest updates that have broken our build. This PR pins us to a specific version of a dependency to work around the problem.

@W-8582993@

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,9 @@
     "tslint": "^5.8.0",
     "typescript": "^4.0.3"
   },
+  "resolutions": {
+    "**/vscode-languageserver-protocol": "3.15.3"
+  },
   "scripts": {
     "build": "lerna run build",
     "clean": "lerna run clean && rimraf yarn.lock && rimraf node_modules",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -22,6 +22,9 @@
     "tslint": "^5.8.0",
     "typescript": "^4.0.3"
   },
+  "resolutions": {
+    "**/vscode-languageserver-protocol": "3.15.3"
+  },
```
jgellin-sf, line 27: Yarn will not automatically hoist the pinned version of a dependency of a dependency, even when it is in the accepted range for resolution. This field overrides yarn's default behavior.

dobladez:
LGTM

dehru (approved):

dehru:
Do we need to publish this?

jgellin-sf:
Probably, but there's no version bump in this PR. My intention was to unblock development, and this is a start.

PR #106, opened 2020-12-16
Disabled 'Running...' button while query running

jgellin-sf:
### What does this PR do?
This PR provides the SOQL Builder UI-side function for indication that a query is running. When the `Run Query` button is pressed, the button is transformed into a disabled `Running...` button, and remains so until VSCode sends the `run_query_done` message.
runquery

@W-8207738@

### DIFF
--- a/packages/soql-builder-ui/src/modules/querybuilder/app/app.test.ts
+++ b/packages/soql-builder-ui/src/modules/querybuilder/app/app.test.ts
@@ -17,7 +17,7 @@ import {
   MessageType,
   SoqlEditorEvent
 } from '../services/message/soqlEditorEvent';
-import { BehaviorSubject, Observable } from 'rxjs';
+import { ArgumentOutOfRangeError, BehaviorSubject, Observable } from 'rxjs';
 import { MessageServiceFactory } from '../services/message/messageServiceFactory';
 import { IMessageService } from '../services/message/iMessageService';
 import { StandaloneMessageService } from '../services/message/standaloneMessageService';
@@ -27,9 +27,9 @@ class TestMessageService implements IMessageService {
   messagesToUI: Observable<SoqlEditorEvent> = new BehaviorSubject(
     ({} as unknown) as SoqlEditorEvent
   );
-  sendMessage() {}
-  setState() {}
-  getState() {}
+  sendMessage() { }
+  setState() { }
+  getState() { }
 }
 
 class TestApp extends App {
@@ -40,6 +40,8 @@ class TestApp extends App {
   isFromLoading = false;
   @api
   isFieldsLoading = false;
+  @api
+  isQueryRunning = false;
   @api
   hasUnrecoverableError = false;
 }
@@ -207,6 +209,14 @@ describe('App should', () => {
         type: MessageType.RUN_SOQL_QUERY
       });
     });
+  });
+
+  it('should clear isQueryRunning flag when run query returns', async () => {
+    app.isQueryRunning = true;
+    messageService.messagesToUI.next({
+      type: MessageType.RUN_SOQL_QUERY_DONE
+    });
+    expect(app.isQueryRunning).toEqual(false);
   });
 
   it('send orderby message to vs code when orderby added', () => {
--- a/packages/soql-builder-ui/src/modules/querybuilder/app/app.test.ts
+++ b/packages/soql-builder-ui/src/modules/querybuilder/app/app.test.ts
@@ -17,7 +17,7 @@ import {
   MessageType,
   SoqlEditorEvent
 } from '../services/message/soqlEditorEvent';
-import { ArgumentOutOfRangeError, BehaviorSubject, Observable } from 'rxjs';
+import { BehaviorSubject, Observable } from 'rxjs';
 import { MessageServiceFactory } from '../services/message/messageServiceFactory';
 import { IMessageService } from '../services/message/iMessageService';
 import { StandaloneMessageService } from '../services/message/standaloneMessageService';
--- a/packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
@@ -109,6 +109,14 @@ button:hover {
     var(--soql-color-blue-text)
   );
 }
+button:disabled {
+  color: var(--vscode-foreground, var(--soql-foreground));
+  background-color: var(
+    --vscode-list-inactiveSelectionBackground,
+    var(--soql-color-light-grey)
+  );
+  cursor: default;
+}
 .select-long {
   min-width: var(--soql-input-width);
   max-width: var(--soql-input-width);
--- a/packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
@@ -109,13 +109,13 @@ button:hover {
     var(--soql-color-blue-text)
   );
 }
-button:disabled {
-  color: var(--vscode-foreground, var(--soql-foreground));
+.btn--disabled {
   background-color: var(
     --vscode-list-inactiveSelectionBackground,
-    var(--soql-color-light-grey)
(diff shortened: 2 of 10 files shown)


packages/soql-builder-ui/src/modules/querybuilder/app/app.test.ts
```diff
@@ -17,7 +17,7 @@ import {
   MessageType,
   SoqlEditorEvent
 } from '../services/message/soqlEditorEvent';
-import { BehaviorSubject, Observable } from 'rxjs';
+import { ArgumentOutOfRangeError, BehaviorSubject, Observable } from 'rxjs';
```
dehru, line 20: Is this one of those sneaky IDE imports?

packages/soql-builder-ui/src/modules/querybuilder/services/toolingSDK.test.ts
```diff
@@ -100,4 +100,13 @@ describe('Tooling SDK Service', () => {
...
     );
   });
+
+  it('Receive run_query_done message', () => {
+    const runStateObserver = jest.fn();
+    toolingSDK.queryRunState.subscribe(runStateObserver);
+    postMessageFromVSCode({
+      type: MessageType.RUN_SOQL_QUERY_DONE
+    });
+    expect(runStateObserver.mock.calls.length).toBe(2);
```
dehru, line 110: Is this 2 because behavior subject gets an immediate callback with the current value?

packages/soql-builder-ui/src/modules/querybuilder/services/message/standaloneMessageService.ts
```diff
@@ -18,6 +18,17 @@ class MockVscode {
   private localStorage = getLocalStorage();
   postMessage(messageObj) {
     this.window.parent.postMessage(messageObj, '*');
+    if (messageObj.type === MessageType.RUN_SOQL_QUERY) {
```
dehru, line 21: we actually do the full round trip across an iframe using this development repo that behaves like vscode.  So you can just recieve the message there and respond using a timeout.

https://github.com/forcedotcom/soql-tooling-standalone/blob/main/src/client/modules/my/app/app.ts#L23

dehru (changes requested):
This UX has bugged me since the start.  Great to get this in.  I had one comment about the setTimeout to mock the response.  Let's move that to the standalone server repo.  So you can literally just remove that whole mocking of the response from this PR.

packages/soql-builder-ui/src/modules/querybuilder/header/header.html
```diff
@@ -8,6 +8,13 @@
 
 <template>
   <header>
-    <button class="run-button" onclick={handleRunQuery}>Run Query</button>
+    <div class="container">
+      <template if:false={isRunning}>
+        <button class="run-button" onclick={handleRunQuery}>Run Query</button>
+      </template>
+      <template if:true={isRunning}>
+        <button class="running-button" disabled>Running...</button>
```
jonnyhork, line 16: Another way to do this without two templates is by making the text of the button dynamic like we do with the [fields](https://github.com/forcedotcom/soql-tooling/blob/f611867098c2fbe360b80967016d8039ebd8e8f7/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts#L65) throughout the UI.
```javascript
    return this.isLoading ? 'Loading...' : 'Select Field...';
```

In this case, if we did not use the templates, the classes and the disabled would require some JS and I think would be more work without that much benefit other than consistency I guess.

I do not see the class `running-button` in the CSS though are you still using it? 
I probably would have used a different class name than `run-button` if I did this again, I'm trying out this whole [BEM](https://css-tricks.com/bem-101/) thing for new work.
For example:
```CSS
header__button--run /* or ready, enabled */
header__button--running /* disabled */

header__button-run--enabled
header__button-run--disabled
```

I am still getting used to it but the idea is to be consistant and predictable with our class names.

jonnyhork (commented):
Looks good for the most part, nice and simple. Just a few comments on the CSS and how we want to approach JS/CSS overall in the UI.

packages/soql-builder-ui/src/modules/querybuilder/app/app.test.ts
jgellin-sf: Yup. Removed with `remove unused import` commit.

packages/soql-builder-ui/src/modules/querybuilder/services/toolingSDK.test.ts
jgellin-sf: Yes

packages/soql-builder-ui/src/modules/querybuilder/header/header.html
jonnyhork: Actually for BEM naming I don't think we include the `header` part... but same idea

packages/soql-builder-ui/src/modules/querybuilder/services/message/standaloneMessageService.ts
jgellin-sf: Ah cool. I've moved it to the tool in this https://github.com/forcedotcom/soql-tooling-standalone/pull/3 and removed from this PR with the `Move standalone run query response to standalone tool` commit.

packages/soql-builder-ui/src/modules/querybuilder/header/header.html
jgellin-sf: It turns out I wasn't styling `run-button` or `running-button` so I've removed those from the template in the `Remove un-styled classes` commit.

packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
```diff
@@ -109,6 +109,14 @@ button:hover {
...
   );
 }
+button:disabled {
+  color: var(--vscode-foreground, var(--soql-foreground));
+  background-color: var(
+    --vscode-list-inactiveSelectionBackground,
+    var(--soql-color-light-grey)
+  );
+  cursor: default;
+}
```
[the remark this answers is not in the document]
jgellin-sf: For the sake of JS ease of use, I think it's probably better to handle disabled button just using CSS as you have done rather than with the attribute. Then it's treated as just another class and not this special case.

I agree it shoud be handled consistently in the CSS common. I see in the where branch the CSS def is in `where.css`. Maybe you can move it to CSS common and we can merge the WHERE stuff, then I'll change to that class when I rebase.

packages/soql-builder-ui/src/modules/querybuilder/header/header.html
jgellin-sf: Whoops acted too fast. `run-button` is used in the test code. Restored now.

packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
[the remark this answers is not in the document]
jonnyhork: ok, well my branch is already merged to `feature/where` but I can create a new PR to move the class to CSS Common, is that what you were thinking?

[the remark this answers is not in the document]
jgellin-sf: Yeah that sounds good.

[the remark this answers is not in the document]
jonnyhork: @jgellin-sf I created and merged #108 to the `feature/where` for your pleasure.

packages/soql-builder-ui/src/modules/querybuilder/header/header.html
dehru: That's relevant as we have discussed moving away ( slowly ) from using css classes as selectors in testing code, because it's often not apparent that the class is for testing vs styling.

jonnyhork: I agree with moving that direction

dehru (approved):

PR #107, opened 2020-12-16
Add basic code-completion for ORDER BY and other improvements

dobladez:
### What does this PR do?

Add basic code-completion for ORDER BY.
Depends on https://github.com/forcedotcom/salesforcedx-vscode/pull/2792

Also:
    * New implementation for extracting data from query (like SObject name),
    * More robust in case of parsing errors (uses custom ANTLR Error Strategy)
    * Better handling of nested queries
    * Better handling of incomplete queries (nested or not)
    * New completions: FOR, OFFSET, LIMIT, ORDER BY, GROUP BY, WITH, WHERE,
                    UPDATE TRACKING, UPDATE VIEWSTAT, COUNT(...), TYPEOF, DISTANCE
    * Refactoring: Extract methods. Move class to new file. Remove old code

soql_orderby_completion1
soql_orderby_completion2

W-8432316

### DIFF
--- a/packages/language-server/src/completion.test.ts
+++ b/packages/language-server/src/completion.test.ts
@@ -278,10 +278,7 @@ describe('Some keyword candidates after FROM clause', () => {
     )
   );
 
-  validateCompletionsFor(
-    'SELECT id FROM Account LIMIT |',
-    expectItems(CompletionItemKind.Constant, '0', '1', '5', '10', '25', '100')
-  );
+  validateCompletionsFor('SELECT id FROM Account LIMIT |', []);
 });
 
 function expectKeywords(...words: string[]): CompletionItem[] {
--- a/packages/language-server/src/completion.test.ts
+++ b/packages/language-server/src/completion.test.ts
@@ -130,17 +130,9 @@ describe('Code Completion on nested select fields: SELECT ... FROM XYZ', () => {
 
 // describe.only('ONLY', () => {
 //   validateCompletionsFor(
-//     'SELECT (SELECT ), | FROM Foo',
-//     sobjectsFieldsFor('Foo')
+//     'SELECT (SELECT |) FROM Foo',
+//     sobjectsFieldsFor('Object')
 //   );
-// validateCompletionsFor(
-//   'SELECT | (SELECT bar FROM Bar) FROM Foo',
-//   sobjectsFieldsFor('Foo')
-// );
-// validateCompletionsFor(
-//   'SELECT (SELECT |) FROM Foo',
-//   sobjectsFieldsFor('Object')
-// );
 // });
 
 describe('Code Completion on SELECT XYZ FROM...', () => {
@@ -242,6 +234,11 @@ describe('Some keyword candidates after FROM clause', () => {
   );
 
   validateCompletionsFor(
+    'SELECT id FROM Account WITH |',
+    expectKeywords('DATA CATEGORY')
+  );
+
+  validateCompletionsFor(
     'SELECT Account.Name, (SELECT FirstName, LastName FROM Contacts |) FROM Account',
     expectKeywords(
       'FOR',
@@ -254,7 +251,7 @@ describe('Some keyword candidates after FROM clause', () => {
       'UPDATE TRACKING',
       'UPDATE VIEWSTAT'
     ),
-    true
+    { skip: true }
   );
 });
 
@@ -269,10 +266,10 @@ function expectKeywords(...words: string[]): CompletionItem[] {
 function validateCompletionsFor(
   text: string,
   expectedItems: CompletionItem[],
-  skip: boolean = false,
-  cursorChar: string = '|'
+  options: { skip?: boolean; only?: boolean; cursorChar?: string } = {}
 ) {
-  const itFn = skip ? xit : it;
+  const itFn = options.skip ? xit : options.only ? it.only : it;
+  const cursorChar = options.cursorChar || '|';
   itFn(text, () => {
     if (text.indexOf(cursorChar) != text.lastIndexOf(cursorChar)) {
       throw new Error(
--- a/packages/language-server/src/completion.test.ts
+++ b/packages/language-server/src/completion.test.ts
@@ -18,11 +18,20 @@ const INNER_SELECT_snippet = {
   insertTextFormat: InsertTextFormat.Snippet,
 };
 
+const COUNT_snippet = {
+  kind: CompletionItemKind.Snippet,
+  label: 'COUNT(...)',
+  insertText: 'COUNT($1)',
+  insertTextFormat: InsertTextFormat.Snippet,
+};
+
 const expectedSimpleFieldCompletions: CompletionItem[] = [
   {
     kind: CompletionItemKind.Field,
     label: '__SOBJECT_FIELDS_PLACEHOLDER:Object',
   },
+  ...expectKeywords('TYPEOF', 'DISTANCE', 'COUNT()'),
+  COUNT_snippet,
   INNER_SELECT_snippet,
 ]; /*.concat(expectKeywords('COUNT')) */
 
@@ -34,6 +43,7 @@ const expectedSObjectCompletions: CompletionItem[] = [
 ];
 
 describe('Code Completion on SELECT ...', () => {
+  validateCompletionsFor('|', expectKeywords('SELECT').concat(SELECT_snippet));
   validateCompletionsFor('SELE|', expectKeywords('SELECT'));
   validateCompletionsFor('SELECT|', expectKeywords('SELECT'));
   validateCompletionsFor('SELECT |', expectedSimpleFieldCompletions);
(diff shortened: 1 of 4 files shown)


packages/language-server/src/completion.test.ts
```diff
@@ -177,22 +219,68 @@ describe('Code Completion on SELECT FROM (no columns on SELECT)', () => {
...
 });
 
-describe('Some keyword candidates', () => {
+describe('Code Completion for ORDER BY', () => {
+  validateCompletionsFor('SELECT id FROM Account ORDER BY |', [
+    {
+      kind: CompletionItemKind.Field,
+      label: '__SOBJECT_FIELDS_PLACEHOLDER:Account',
+    },
+    ...expectKeywords('DISTANCE'),
```
praksb, line 237: why do you also expect DISTANCE here?

packages/language-server/src/completion.test.ts
```diff
@@ -177,22 +219,68 @@ describe('Code Completion on SELECT FROM (no columns on SELECT)', () => {
...
+  ]);
+});
+
+describe('Some keyword candidates after FROM clause', () => {
   validateCompletionsFor(
     'SELECT id FROM Account |',
-    expectKeywords('LIMIT', 'WHERE', 'ORDER BY', 'GROUP BY')
+    expectKeywords(
+      'FOR',
```
praksb, line 245: Just curious why you don't have "FOR VIEW" and "FOR REFERENCE" here similar to UPDATE TRACKING and UPDATE VIEWSTAT?

packages/language-server/src/completion.test.ts
```diff
@@ -177,22 +219,68 @@ describe('Code Completion on SELECT FROM (no columns on SELECT)', () => {
...
+      'WITH',
+      'WHERE',
+      'UPDATE TRACKING',
+      'UPDATE VIEWSTAT'
+    )
+  );
+
+  validateCompletionsFor(
+    'SELECT id FROM Account LIMIT |',
+    expectItems(CompletionItemKind.Constant, '0', '1', '5', '10', '25', '100')
```
praksb, line 283: are you using this for testing? I don't think we should provide these as part of auto-completion for LIMIT clause. This can take any integer value right not just the ones that you have called out?

packages/language-server/src/completion/SoqlCompletionErrorStrategy.ts
```diff
@@ -0,0 +1,19 @@
...
+    // The default error handling strategy is "too smart" for our code-completion purposes.
+    // We generally do NOT want the parser to remove tokens for recovery.
+    //
+    // Example:
+    //    SELECT id, | FROM Foo
+    //
+    // Here the parser drops `FROM` and makes Foo a field on SELECTs' list
+    // So we don't recognize Foo as the SObject we want to query for.
+    //
+    // We might implement more SOQL-specific logic in the future.
```
praksb, line 16: It would be nice to capture these comments in a doc comment format for typescript. It looks like there is a proposal here to standardize doc comments: https://tsdoc.org/. Can we move this comment to follow that?

packages/language-server/src/completion/SoqlQueryExtractor.ts
```diff
@@ -0,0 +1,109 @@
...
+  matchedSelectFroms: Array<InnerSoqlQuery> = [];
+
+  innerSoqlQueries = new Map<number, InnerSoqlQuery>();
+
+  constructor(private cursorTokenIndex: number) {}
+
+  static getSObjectFor(
+    parsedQueryTree: SoqlQueryContext,
+    cursorTokenIndex: number
+  ): string | undefined {
```
praksb, line 36: I wonder if we need to add specific unit tests to test the code in this class. Specifically tests that verify getSObjectFor() function directly.

packages/language-server/src/completion.ts
```diff
@@ -176,60 +156,150 @@ export function findCursorTokenIndex(
...
+            completionItems.push(newSnippetItem('COUNT(...)', 'COUNT($1)'));
+            completionItems.push(
+              newSnippetItem('(SELECT ... FROM ...)', '(SELECT $2 FROM $1)')
+            );
+          }
+        }
+        break;
+      case SoqlParser.RULE_soqlInteger:
+        if (isCursorAfter(tokenStream, tokenIndex, [SoqlLexer.LIMIT])) {
+          completionItems.push(newNumberItem('0'));
```
praksb, line 246: I don't think showing completion here is necessary.

packages/language-server/src/completion.ts
```diff
@@ -176,60 +156,150 @@ export function findCursorTokenIndex(
...
+): CompletionItem[] {
+  const completionItems: CompletionItem[] = [];
+
+  for (let [ruleId, ruleData] of c3Rules) {
+    switch (ruleId) {
+      case SoqlParser.RULE_soqlUpdateStatsClause:
+        // NOTE: We handle this one as a Rule instead of Tokens because
+        // "TRACKING" and "VIEWSTAT" are not part of the grammar
+        if (tokenIndex == ruleData.startTokenIndex) {
+          completionItems.push(newKeywordItem('UPDATE TRACKING'));
```
praksb, line 214: Can you extract this and the string below into constants?

packages/language-server/src/completion.test.ts
dobladez: It's one of the standard functions which can be used in SELECT, WHERE, ORDER BY expressions. It happens that `DISTANCE` is modeled explicitly in the grammar, and so we get its completion from it "automatically".

We should add other standard functions to our completion suggestions. Most are not modeled as part of the grammar.

packages/language-server/src/completion.test.ts
dobladez: Good question.

The completion of `FOR`, followed by `VIEW` or `REFERENCE` comes directly from the grammar tokens. No explicit coding was added for them. The generic code joins a chain of contiguous keywords together only if they are the only next option (example: `ORDER BY`). In the case of `FOR`, it doesn't automatically chain it because there's more than one following option option.

However, for the case of `UPDATE`, we've added an explicit case to handle it because `TRACKING` and `VIEWSTAT` are not modeled as part of the grammar. So, the completion token inferred from the grammar is only `UPDATE`, and nothing else afterwards.

So: if we want to provide `FOR VIEW` and `FOR REFERENCE` a single options, we can easy do it by adding an explicit case for handling them. I think it's a good idea.  The some goes for `NULLS FIRST` and `NULLS LAST` for example.  It'd be nice if C3 provided additional information about all possible next tokens to generalize these cases... enhancing C3 would be another option.

packages/language-server/src/completion.test.ts
dobladez: Not just for testing. We are providing some common values for completion.

packages/language-server/src/completion/SoqlCompletionErrorStrategy.ts
dobladez: Sure

packages/language-server/src/completion.ts
dobladez: Sure

packages/language-server/src/completion/SoqlQueryExtractor.ts
dobladez: IMHO, the higher-level test cases we have should cover the functionality. And if they don't, we must add new cases.

What I don't like about such lower-level tests is that they are brittle: they break and the minimal refactoring. For instance: I already changed `SoqlQueryExtractor` (on a separate branch) to receive `parsedQueryTree` in the constructor instead... this change would break such unit tests.

**That said**, there are code-paths which cannot be covered from higher-level tests. For example, line 50 below, which Codecov reports as untested: given the grammar definition, it's impossible to have an `innerQuery` with an undefined `.parent`. To test this scenario, we have to cook an artificial parse tree which cannot be done from higher level tests. I will try to do it with the next branch/PR.

packages/language-server/src/completion.ts
dobladez: OK. You can see it in action at the end of the first video on the PR description. I'll remove it if you don't like it  :'(

packages/language-server/src/completion.test.ts
praksb: Thanks for clarifying.

packages/language-server/src/completion/SoqlQueryExtractor.ts
praksb: You have some good points. My main concern was redundancy in higher-level tests where we may end up adding more tests than needed across the various clauses and that is not necessarily bad.

packages/language-server/src/completion.ts
praksb: Can we remove it? The pre-populated list seems arbitrary and worry if it may lead to customer confusion.

dobladez: Done.

praksb (approved):

PR #109, opened 2020-12-22
Type-aware where

jgellin-sf:
### What does this PR do?
This PR:
- creates a utility to get a field type from SObject metadata
- creates a validation framework for where input using a factory pattern
- implements validation for boolean, currency, date, integer, floating point, and pickist types
- implements normalization (quoting and escaping of string values
- implements special case normalization for null input
- shows an error message when validation of where input fails

This PR does not:
- show an error message in a beautiful elegant manner
- change the HTML input based on type

@W-8555377@


packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.ts
```diff
@@ -0,0 +1,67 @@
...
+  @api selectedIndex = '-1';
+
+  renderedCallback() {
+    const buttonElements = this.template.querySelectorAll('button');
+    buttonElements.forEach(element => {
+      this.applyStyle(element);
+    });
+  }
+
+  get buttonData() {
```
dehru, line 23: This would be best to in the api.  so add get/set to the buttonLabels and do it in the setter.  Then just render buttonData.  It's best to keep as much looping as possible outside of the render loop of LWC, since we don't know how often it attempts to re-render itself.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+import { api, LightningElement, track } from 'lwc';
+
+export default class ButtonGroup extends LightningElement {
+  @api buttonLabels: string[];
+  @api selectedIndex = '-1';
+
+  renderedCallback() {
+    const buttonElements = this.template.querySelectorAll('button');
```
dehru: Opposite of my last comment, this could be best to do in the render loop from html.  You can call a function from the template with `class={applyStyle(index)} ` in the template.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.ts
dehru: You could also apply the class names in a function called from the template.
`class={getClassName(idx)}`
This would allow the button to dynamically restyle whenever the selectedIndex is changed.  Altering the selected index will trigger a re-render and calculate the new styles...the most important being which is selected.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.ts
jgellin-sf: Would love to do this, but you can't have a call with arguments in the template.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.css
```diff
@@ -0,0 +1,55 @@
...
+ *  For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ *
+ */
+@import 'querybuilder/cssCommon';
+
+.btn--group {
+  align-self: flex-start;
+}
+.btn {
+  height: 1.5em;
```
dehru, line 14: I generally like to use `rem` vs `em` personally.  Because em is dynamic as compared to it's parent.  So that can be funky as text up the dom possibly gets set to "larger" or "smaller", it affects what an em is on the button.  Where as rem is based on the root font size, which is just easier to think about logically for me.

Normally, it's not a big issue, but i wanted to mention it from my experience.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    expect(buttonGroup.selection).toEqual(1);
+  });
+
+  it('fire selection changed event when button clicked', () => {
+    buttonGroup.selectedIndex = '1';
+    const handler = jest.fn();
+    buttonGroup.addEventListener('selection__changed', handler);
+    document.body.appendChild(buttonGroup);
+
+    const buttonElements = buttonGroup.shadowRoot.querySelectorAll('button');
```
dehru, line 60: I think you can just do
`const button0 = buttonGroup.shadowRoot.querySelectorAll('button')[0];`

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+  isLast(index: number): boolean {
+    return index === this.buttonLabels.length - 1;
+  }
+  isSelected(index: number): boolean {
+    return index === parseInt(this.selectedIndex);
+  }
+
+  @api get selection() {
+    return parseInt(this.selectedIndex);
```
dehru, line 50: How do you see this being used vs the event?  I noticed you are using it in testing, however, outside of testing, i think the event will likely drive the app.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.ts
jgellin-sf: Ok I put the className in the buttonData so everything needed to render the template is in that object. Then on selection I update the style. Since the buttonData is tracked LWC knows to re-render.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.ts
jgellin-sf: Sorta kinda do this now with keeping the classlist in the buttondata.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.css
jgellin-sf: I like that better. I made the change.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.test.ts
jgellin-sf: Thanks. Done.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.ts
jgellin-sf: I agree. It's gone now.

packages/soql-builder-ui/src/modules/querybuilder/buttonGroup/buttonGroup.ts
dehru: Yeah.  the limitations of LWC often catch me as a surprise coming from Angular / React's much richer template syntax.

packages/soql-model/src/messages/messages.ts
```diff
@@ -12,4 +12,13 @@ export namespace Messages {
   export const error_noFrom = 'A FROM clause is required.';
   export const error_incompleteFrom = 'Incomplete FROM clause. The FROM clause requires an object.';
   export const error_incompleteLimit = 'Incomplete LIMIT clause. The LIMIT keyword must be followed by a number.';
+
+  export const error_fieldInput_boolean = 'Value must be TRUE or FALSE';
```
ekapner, line 16: Does case matter here? Must be all caps?

packages/soql-model/src/messages/messages.ts
```diff
@@ -12,4 +12,13 @@ export namespace Messages {
...
+  export const error_fieldInput_currency = 'Currency value is not valid';
```
ekapner, line 17: Any guidelines we can add here?

packages/soql-model/src/messages/messages.ts
```diff
@@ -12,4 +12,13 @@ export namespace Messages {
...
+  export const error_fieldInput_date = 'Date value is not valid';
```
ekapner, line 18: Can we provide valid date format. like: You must enter the date as YYYY-MM-DD (or something like that)

packages/soql-model/src/messages/messages.ts
```diff
@@ -12,4 +12,13 @@ export namespace Messages {
...
+  export const error_fieldInput_float = 'Value must be numeric';
```
ekapner, line 19: Does this mean "numbers only"?  decimals?

packages/soql-model/src/messages/messages.ts
```diff
@@ -12,4 +12,13 @@ export namespace Messages {
...
+  export const error_fieldInput_integer = 'Value must be a whole number';
+  export const error_fieldInput_picklist = 'Value must be one of: {0}';
+
+  export const error_operatorInput = '{0} operator cannot be used for this field type';
```
ekapner, line 23:
```suggestion
  export const error_operatorInput = '{0} operator can't be used for this field type';
```

ekapner (commented):
A couple questions before I can finalize the validation messages

packages/soql-model/src/messages/messages.ts
jgellin-sf: Case does not actually matter, but all-caps is consistent with the way the documentation (https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql.htm) writes elements of SOQL syntax.

packages/soql-model/src/messages/messages.ts
jgellin-sf: There's a 3 letter currency code followed by the numeric value according to the [ISO standard](https://en.wikipedia.org/wiki/ISO_4217). So $100 US is USD100. I _think_ the currency value can also be just a floating point number. So the most concise we can get is maybe...

"Value must be a number or a three-letter ISO currency code followed by a number"

packages/soql-model/src/messages/messages.ts
jgellin-sf: Date is tricky. Because you can write date as:
- Date only: 2021-01-13
- UTC date and time: 2021-01-13T12:00:00Z
- Date and time with timezone offset: 2021-01-13T12:00:00-05:00
- Any of the many range literals built into SOQL (TODAY, TOMORROW, NEXT_WEEK, LAST_YEAR, etc)
- Any of the many parameterized date range literals build into SOQL (NEXT_N_DAYS:5, LAST_N_YEARS:20, etc.)
See https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_dateformats.htm?search_text=currency

packages/soql-model/src/messages/messages.ts
jgellin-sf: Right. So numeric values, containing decimal or not, are fine.

packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
```diff
@@ -38,6 +39,15 @@ export default class App extends LightningElement {
   modelService: ToolingModelService;
   messageService: IMessageService;
   theme = 'light';
+  _sobjectTypeUtils: SObjectTypeUtils;
```
dehru, line 42: Is there a reason to use a getter/setter here?

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.test.ts
```diff
@@ -168,4 +169,71 @@ describe('WhereModifierGroup should', () => {
...
+    });
+    let resultingCriteria;
+    const handler = (e => {
+      resultingCriteria = e.detail.criteria;
+    });
+    modifierGroup.addEventListener('modifiergroupselection', handler);
+    document.body.appendChild(modifierGroup);
+
+    const { criteriaInputEl } = getModifierElements();
+    criteriaInputEl.value = 'FALSE';
```
dehru, line 234: would this test pass if criteriaInputEl.value was lower case 'false'?

packages/soql-model/src/messages/messages.ts
dehru: I feel like including the `three-letter ISO currency code` makes this feel more complicated than necessary.

packages/soql-model/src/validators/integerValidator.ts
```diff
@@ -0,0 +1,16 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { Messages } from '../messages/messages';
+import { ValidateResult, Validator } from './validator';
+
+export class IntegerValidator extends Validator {
+  public validate(input: string): ValidateResult {
+    const isValid = /^[+-]?[0-9]+$/.test(input.trim());
```
dehru, line 12: There's also Number.isInteger() or parseInt().  But i don't know if they are more inclusive than you want.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -106,38 +113,135 @@ export default class WhereModifierGroup extends LightningElement {
...
+        type: critType,
+        value: normalizedInput
+      };
+      this.errorMessage = '';
+
+      const validateOptions = {
+        type,
+        picklistValues
+      };
+
```
dehru, line 214: I notice we are checking 2 validations here, but only displaying 1 error message.  In your experience with this, is it best to check the input validation for the field first, and if that passes, show check operator validation?

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -5,22 +5,29 @@
...
+import { displayValueToSoqlStringLiteral, soqlStringLiteralToDisplayValue } from '../services/soqlUtils';
+
 
 export default class WhereModifierGroup extends LightningElement {
   @api allFields: string[];
   @api selectedField: string = undefined;
   @api selectedOperator: string;
   @api isLoading = false;
   @api index;
+  @api sobjectTypeUtils: SObjectTypeUtils;
```
dehru, line 23: I'm thinking we should pass the metadata and then process it here instead of with app.ts.  using a setter to process the data with sobjectutils

packages/soql-builder-ui/src/modules/querybuilder/services/sobjectUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+ *  All rights reserved.
+ *  Licensed under the BSD 3-Clause license.
+ *  For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ *
+ */
+
+import { Soql } from '@salesforce/soql-model';
+
+export class SObjectTypeUtils {
+  constructor(protected sobjectMetadata: any) {
```
dehru, line 12: To limit the looping/filtering in this utility, we could use a map so that the key is the field name ( possibly lowercased ).  Something like this:
```
{
    'id' : { name: 'Id', type: 'id', picklistValues: [] },
    'name' : { name: 'Name', type: 'string', picklistValues: [] },
    'accountsource': { name: 'AccountSource', type: 'picklist', picklistValues: [{ value: 'apple' }, { value: 'banana' }, { value: 'cherry' }] },
      'annualrevenue': { name: 'AnnualRevenue', type: 'currency', picklistValues: [] },
      'billingaddress' : { name: 'BillingAddress', type: 'address', picklistValues: [] },
       ...
}
```

dehru:
This is amazingly cool work JG!

dehru (approved):
None of my comments are change requests.  Just possible enhancements and/or thinking out loud.  Address any you see fit.  

We could even merge this so that I can attempt to beautify the message while you address any of the comments.

packages/soql-model/src/messages/messages.ts
ekapner: They are all a bit complicated - so let's just leave them as @jgellin-sf wrote them. Thanks for answering my questions in such detail.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.test.ts
jgellin-sf: yes

packages/soql-model/src/validators/integerValidator.ts
jgellin-sf: Those will return a integer for any string starting with an integer, so parseInt() will return 134 for " 134hello".

packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
jgellin-sf: I don't think it's really necessary.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
jgellin-sf: I don't have any special insight to the order of the validations. It felt right. 😂

packages/soql-builder-ui/src/modules/querybuilder/services/sobjectUtils.ts
jgellin-sf: I updated this code in the `Refactor sobject utils for easier map lookups` commit.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
jgellin-sf: Made this change in `pass sobject metadata rather than utils object` commit

packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
jgellin-sf: This is irrelevant following the `pass sobject metadata rather than utils object` commit

packages/soql-model/src/validators/integerValidator.ts
dehru: yeah, i guess that's a false positive.

PR #111, opened 2020-12-23
Jhork/custom select fields

jonnyhork:
### What does this PR do?
This PR will add a new `<custom-select>` custom select component and use it in the SELECT section of the SOQL Builder.(selecting fields from an object)

Custom-Select Demo (dark)

**Expected Behavior**
- The user can click on the search bar to see all the available options that are not already selected
- The user can type to search for a option
- If the user types a valid field name and hits 'enter' the option will be selected
- The user can navigate the list of options with the arrow keys
- The user can select and option that is highlighted with 'enter'
- The user can clear the search bar with the "X"

**API**
The component is designed to be generalized so that is can more easily be used in other parts of the app.

`placeholder-text`: The value of this attribute will be displayed as the placeholder of the search bar when is not loading.
`option__selection`: This is the name of the event dispatched when an option is selected, to be handled by the parent.
```javascript
// event payload
{
 detail: {
   optionValue: string
  }
}
```
**Proposed Features**
- Add all options
- Remove all options 
- True multi-select UX that allows users to select more than one option at a time
- Autocomplete while typing in the search bar
@W-7987352@

### DIFF
--- a/packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
+++ b/packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
@@ -46,7 +46,7 @@ export default class App extends LightningElement {
   }
 
   get hasUnrecoverable() {
-    return !this.hasUnsupported && this.hasUnrecoverableError
+    return !this.hasUnsupported && this.hasUnrecoverableError;
   }
 
   get blockQueryBuilder() {
@@ -88,7 +88,7 @@ export default class App extends LightningElement {
       this.isFieldsLoading = false;
       this.fields =
         sobjectMetadata && sobjectMetadata.fields
-          ? sobjectMetadata.fields.map((f) => f.name)
+          ? sobjectMetadata.fields.map((f) => f.name).sort()
           : [];
     });
     this.loadSObjectDefinitions();
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
@@ -1,0 +1,84 @@
+@import 'querybuilder/cssCommon';
+
+/* ======= SELECT ======= */
+.select__wrapper {
+  display: flex;
+  justify-content: space-between;
+  position: relative;
+  width: 100%;
+  height: 1.5rem;
+  border: 1px solid
+    var(--vscode-list-inactiveSelectionBackground, var(--soql-color-light-grey));
+  border-radius: 3px;
+  background-color: var(--vscode-sideBar-background, var(--soql-background));
+}
+/* .select__wrapper::before {
+  position: absolute;
+  content: '🔍';
+  font-size: 1em;
+  top: 1px;
+  padding-left: 5px;
+  background-color: var(--vscode-sideBar-background, var(--soql-background));
+} */
+
+.select__input {
+  width: 90%;
+  height: 1em;
+  padding: 4px;
+  border: none;
+}
+.select:focus {
+  outline: none;
+}
+
+.select__clear-search {
+  /* font-size: 1em; */
+  align-self: center;
+  padding-right: 5px;
+  cursor: pointer;
+}
+
+/* ======== OPTIONS ======== */
+
+.options__wrapper {
+  position: absolute;
+  top: 64px;
+  border: 1px solid
+    var(--vscode-list-inactiveSelectionBackground, var(--soql-color-light-grey));
+  border-radius: 3px;
+  width: var(--soql-input-width);
+  min-height: 20%;
+  max-height: 50%;
+  overflow-y: scroll;
+  /* hidden by default */
+  visibility: hidden;
+  opacity: 0;
+  transition: opacity 0.2s;
+  pointer-events: none;
+  z-index: 200;
+  background-color: var(--vscode-sideBar-background, var(--soql-background));
+}
+
+.options--open {
+  opacity: 1;
+  visibility: visible;
+  pointer-events: all;
+}
+
+.option {
+  padding-left: 5px;
+}
+.option:hover {
+  background-color: #4dd0e130;
+}
+.option:focus {
+  background-color: #4dd0e130;
+}
+.option--highlight {
+  background-color: #4dd0e130;
+}
+
+.misc-input {
+  margin-top: 20px;
+  width: 90%;
+}
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
@@ -25,7 +25,8 @@
   left: 6px;
   transform: rotate(45deg);
   border-radius: 50%;
-  background-color: var(--vscode-sideBar-background, var(--soql-background));
+  background-color: transparent;
+  vertical-align:bottom;
 }
 
 .select__input {
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
@@ -93,3 +93,11 @@
     var(--soql-color-medium-grey)
   );
 }
+.option--placeholder {
+  color: var(
+    --vscode-list-inactiveFocusBackground,
+    var(--soql-color-medium-grey)
+  );
+  cursor: not-allowed;
+  pointer-events: none;
+}
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
@@ -12,21 +12,26 @@
   border-radius: 3px;
   background-color: var(--vscode-sideBar-background, var(--soql-background));
 }
-/* .select__wrapper::before {
+/* rotate nueter symbol to mimic search icon */
+.select__wrapper::before {
   position: absolute;
-  content: '🔍';
-  font-size: 1em;
-  top: 1px;
-  padding-left: 5px;
+  content: '⚲';
+  font-size: 1.5em;
+  bottom: -2px;
+  left: 6px;
+  transform: rotate(45deg);
+  border-radius: 50%;
   background-color: var(--vscode-sideBar-background, var(--soql-background));
-} */
+}
 
 .select__input {
   width: 90%;
   height: 1em;
   padding: 4px;
+  margin-left: 24px;
   border: none;
 }
+
 .select:focus {
   outline: none;
 }
@@ -35,6 +40,10 @@
   align-self: center;
   padding-right: 5px;
   cursor: pointer;
+}
+
+.select__arrow--down {
+  font-size: 1em;
 }
 
 /* ======== OPTIONS ======== */
@@ -65,19 +74,18 @@
 }
 
 .option {
-  padding-left: 5px;
+  padding: 3px 0 3px 4px;
+  margin: 0;
 }
 .option:hover {
-  background-color: #4dd0e130;
-}
-.option:focus {
-  background-color: #4dd0e130;
+  background-color: var(
+    --vscode-list-focusBackground,
+    var(--soql-color-medium-grey)
+  );
 }
 .option--highlight {
-  background-color: #4dd0e130;
-}
-
-.misc-input {
-  margin-top: 20px;
-  width: 90%;
+  background-color: var(
+    --vscode-list-focusBackground,
+    var(--soql-color-medium-grey)
+  );
 }
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
@@ -46,10 +46,6 @@
   cursor: pointer;
 }
 
-.select__arrow--down {
-  font-size: 1em;
-}
-
 /* ======== OPTIONS ======== */
 
 .options__wrapper {
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
@@ -63,7 +63,7 @@
   max-height: 50%;
   overflow-y: scroll;
   /* hidden by default */
-  visibility: hidden;
+  display: none;
   opacity: 0;
   transition: opacity 0.2s;
   pointer-events: none;
@@ -73,7 +73,7 @@
 
 .options--open {
   opacity: 1;
-  visibility: visible;
+  display: block;
   pointer-events: all;
 }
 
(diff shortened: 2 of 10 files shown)


packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.html
```diff
@@ -0,0 +1,39 @@
...
+        key={optionName}
+        class="option"
+        data-option-value={optionName}
+      >
+        {optionName}
+      </p>
+    </template>
+    <template if:true={noResultsFound}>
+      <!-- i18n -->
+      <p class="option option--placeholder">No results found.</p>
```
jonnyhork, line 36: @ekapner This text will be displayed to the user when they type in a search term and there are no results. I used the same phrase as VS Code when you search you code base.
Please see the gif in the description for an example.

packages/soql-builder-ui/src/modules/querybuilder/fields/fields.ts
```diff
@@ -12,18 +12,13 @@ export default class Fields extends LightningElement {
   @api selectedFields: string[] = [];
   @api hasError = false;
   @api isLoading = false;
-
-  get defaultOptionText() {
-    // TODO: i18n
-    return this.isLoading ? 'Loading...' : 'Select fields...';
-  }
+  selectPlaceHolderText = 'Search fields...'; // TODO: i18n
```
jonnyhork, line 15: @ekapner The default select elements say `Select fields...` so I just changed it to `Search fields...` to indicate they can type to search.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,265 @@
...
+      this.resetSearchBar();
+    }
+  }
+
+  haveOptionsToNavigate(): boolean {
+    return (
+      !!(!this.optionListIsHidden && this.optionList.length) &&
+      !this.noResultsFound
+    );
+  }
```
jonnyhork, line 104: This is used to control whether the component responds to some keyboard events like arrow navigation.
**If**
- the options list is visible
- there are options to choose from
- There are results of a user search

**then allow keyboard navigation**
It reads a bit odd but the names of the variables make more sense for how they are used in other cases. 

`!!(!this.optionListIsHidden && this.optionList.length)` would return the number value of the length without the double bang, I thought it makes more sense to have a boolean value returned.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  renderedCallback() {
+    this.optionsWrapper = this.template.querySelector('.options__wrapper');
+    this.optionList = this.optionsWrapper.children;
+  }
+
+  /* ======= UTILITIES ======= */
+
+  getAvailableOptions() {
```
jonnyhork, line 52: Sets the available options in state.

available options = `all options - selected options` case insensitive

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,265 @@
...
+  activeOptionIndex = -1;
+  numberOfSearchResults;
+
+  get hasSearchTerm() {
+    return !!this.searchTerm;
+  }
+
+  get noResultsFound() {
+    return this.hasSearchTerm && this.numberOfSearchResults === 0;
+  }
```
jonnyhork, line 24: This is used to render the option placeholder template "No results found." when there is no match to the user search.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,265 @@
...
+
+export default class CustomSelect extends LightningElement {
+  @api isLoading = false;
+  @api allOptions: string[];
+  @api selectedOptions: string[] = [];
+  @track _renderedOptions: string[] = [];
+  availableOptions: string[] = [];
+  _placeholderText = '';
+  searchTerm = '';
+  originalUserInput = '';
```
jonnyhork, line 11: `originalUserInput` is used as a cache for what is typed into the input element, if the user types "na" and then navigates down to a the list of options to "name" the search term becomes "name". 
Then if they navigate back up to the search bar the value becomes "na" again or whatever they typed initially.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  /* ======= EVENT HANDLERS ======= */
+
+  // called when user clicks on search bar input
+  handleOpenOptions(e) {
+    e.preventDefault();
+    e.stopPropagation();
+    this.getAvailableOptions();
+    if (this.hasSearchTerm) {
+      this.filterOptionsBySearchTerm();
```
jonnyhork, line 146: If there is already some text in the search bar and they click on it to see the options list again, it will resemble those that match the search term. 
Otherwise it will be all available options.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  set placeholderText(text: string) {
+    this._placeholderText = text;
+  }
+
+  /* ======= LIFECYCLE HOOKS ======= */
+
+  // close the options menu when user click outside element
+  connectedCallback() {
+    document.addEventListener('click', () => {
```
dehru, line 40: We need to adjust this a bit so that we can also call document.removeEventListener on the disconnectedCallback.

https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.create_lifecycle_hooks_dom

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,265 @@
...
+      this.handleCloseOptions();
+    });
+  }
+
+  renderedCallback() {
+    this.optionsWrapper = this.template.querySelector('.options__wrapper');
```
dehru, line 46: We should check to see if this is already set, before we query again since renderedCallback can happen more than once.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
dehru: Semantically, this function behaves more like `calculateAvailableOptions` since the work "get" makes me think it should return something.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,265 @@
...
+        detail: {
+          optionValue
+        }
+      });
+      this.dispatchEvent(optionSelectionEvent);
+      this.resetSearchBar();
+    }
+  }
+
+  haveOptionsToNavigate(): boolean {
```
dehru, line 99: usually, then verb _tense_ i see here is `hasOptionsToNavigate`

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,265 @@
...
+      this.optionList[this.activeOptionIndex].classList.remove(
+        'option--highlight'
+      );
+    }
+  }
+
+  addOptionHighlight(position: number) {
+    if (this.optionList[position]) {
+      this.optionList[position].classList.add('option--highlight');
+      this.optionList[position].scrollIntoView({
```
dehru, line 117: so nice.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,265 @@
...
+      (option) => {
+        return option.toLowerCase() === optionName.toLowerCase();
+      }
+    );
+
+    if (validOptionMatch.length) {
+      const optionValue = validOptionMatch[0];
+      const optionSelectionEvent = new CustomEvent('option__selection', {
+        detail: {
+          optionValue
```
dehru, line 91: let's just simplify this to `value`

packages/soql-builder-ui/src/modules/querybuilder/fields/fields.ts
```diff
@@ -12,18 +12,13 @@ export default class Fields extends LightningElement {
...
 
   handleFieldSelection(e) {
     e.preventDefault();
-    const value = e.target.value;
-    if (value && value.length) {
+    if (e.detail && e.detail.optionValue) {
       const fieldSelectedEvent = new CustomEvent('fieldselected', {
-        detail: { field: e.target.value }
+        detail: { field: e.detail.optionValue }
```
dehru, line 21: as noted above, seems like `value` is simpler.

dehru (changes requested):
I've got a couple minor comments.  Most are just opinions about naming or stylistic suggestions.  The one we do need to handle though is the `disconnectedCallback` so we can unregister the document click listener.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
```diff
@@ -0,0 +1,451 @@
...
+        })
+        .then(() => {
+          optionsList = customSelect.shadowRoot.querySelector(
+            '.options__wrapper'
+          );
+          expect(optionsList.getAttribute('aria-hidden')).toBe('true');
+        });
+    });
+
+    it('should reset the search bar when X is clicked', () => {
```
dehru, line 114: This test is fantastic.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
```diff
@@ -0,0 +1,451 @@
+/*
```
dehru, line 1: It'd be good to assign all the duplicated class names and such to variables inside the outer `describe` section to prevent duplication.  Like `searchBarSelector = 'input[name=search-bar]'`, or maybe write a couple functions to get those for you each time.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
dehru: disconnectedCallback added

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
dehru: cached and used if set

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
dehru: updated

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
dehru: done

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
dehru: Done

dehru (approved):

dehru:
Let's push this into a feature/custom-select and continue to implement this for the rest of the fields.  This is so cool that i can't face another standard select dropdown.

packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
```diff
@@ -88,7 +88,7 @@ export default class App extends LightningElement {
       this.isFieldsLoading = false;
       this.fields =
         sobjectMetadata && sobjectMetadata.fields
-          ? sobjectMetadata.fields.map((f) => f.name)
+          ? sobjectMetadata.fields.map((f) => f.name).sort()
```
dehru, line 91: Let's sort the fields alphabetically;  I wanted to do it here so that the custom-select component doesn't do any sorting and displays the fields in the order it was given.  It shows it's options in the order given.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
```diff
@@ -0,0 +1,103 @@
...
+  height: 1.5rem;
+  border: 1px solid
+    var(--vscode-list-inactiveSelectionBackground, var(--soql-color-light-grey));
+  border-radius: 3px;
+  background-color: var(--vscode-sideBar-background, var(--soql-background));
+}
+/*
+- rotate nueter symbol to mimic search icon
+- these styles are set for vs code, alignment is off in standalone
+*/
```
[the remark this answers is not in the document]
dehru: I was able to fix this by removing attributes.  I just lucked out.  https://github.com/forcedotcom/soql-tooling/pull/111/commits/f27d36a10ca783d07e211788092828ad276de753

Screen Shot 2021-01-14 at 11 00 29 AM

Screen Shot 2021-01-14 at 11 02 50 AM

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,277 @@
...
+      this.resetSearchBar();
+    }
+  }
+
+  hasOptionsToNavigate(): boolean {
+    return (
+      this.optionListIsHidden === false &&
+      this.optionList.length > 0 &&
+      this.noResultsFound === false
+    );
```
dehru, line 108: I simplified this for readability.  It was a series of `!!(! condition & condition) & !condition)` and I personally was struggling to read it.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -0,0 +1,277 @@
...
+  handleOpenOptions(e) {
+    e.preventDefault();
+    e.stopPropagation();
+    this.calculateAvailableOptions();
+    if (this.hasSearchTerm) {
+      this.filterOptionsBySearchTerm();
+    } else {
+      this._renderedOptions = this.availableOptions;
+    }
+    this.openOptionsMenu();
```
dehru, line 158: I moved this `openOptionsMenu()` to the bottom.  For some reason, if I matched a search term like "Id" but didn't click away from the input box, just kept typing with another search term, this previously selected search term would remain in the options list.  By waiting until last to display it, LWC seems to render the option list correctly.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
dehru: I struggled comprehending this and updated it for readability.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
dehru: I updated this.

PR #113, opened 2021-01-07
Disable WHERE in model

jgellin-sf:
### What does this PR do?
This PR disables WHERE in the soql-model package since the UI is not ready yet.

@W-8657716@


dehru:
This would have taken me a lot longer to figure out!  When the UI is ready, we can just revert this commit.

This passed QA for me.

dehru (approved):

PR #116, opened 2021-01-14
Recognize recoverable where errors

jgellin-sf:
### What does this PR do?
This PR improves the model to identify recoverable errors in the WHERE clause. The intention is to cover all recoverable errors that are supported in the targeted where feature. Recognized errors include:
- empty where clause ("WHERE")
- incomplete nested clause AKA missing paren ("WHERE (field = 5")
- incomplete AND/OR condition ("WHERE field = 5 AND" or "WHERE OR field = 5")
- incomplete NOT condition ("WHERE NOT")
- unrecognized literal value in comparison condition ("WHERE field = foo")
- unrecognized operator in comparison condition ("WHERE field LIK 'foo')
- unrecognized field in comparison condition ("WHERE 5 = 5")
- no literal value in comparison condition ("WHERE field =")
- no operator in comparison condition ("WHERE field")

This PR is for the model only, it does not contain the actual recovery from recognized errors in the UI.

@W-8555673@


packages/soql-model/src/serialization/deserializer.test.ts
```diff
@@ -662,4 +662,100 @@ describe('ModelDeserializer should', () => {
...
+    const expectedType = ErrorType.NOCOMPAREVALUE;
+    const model = new ModelDeserializer('SELECT field1 FROM object1 WHERE field =').deserialize();
+    if (model.errors && model.errors.length === 1) {
+      expect(model.errors[0].type).toEqual(expectedType);
+    } else {
+      fail();
+    }
+  });
+
+  it('identify missing compare operator in condition', () => {
```
dobladez, line 752: These test cases seem to all have the same structure. How about a helper function? Then the test cases could look something like:
```
    expectError('SELECT field1 FROM object1 WHERE field =', ErrorType.NOCOMPAREVALUE));
    expectError('SELECT field1 FROM object1 WHERE 5 = 5', ErrorType.UNRECOGNIZEDCOMPAREFIELD));
    ...etc...
```

packages/soql-model/src/serialization/deserializer.ts
```diff
@@ -113,6 +115,78 @@ class ErrorIdentifier {
         charInLine: error.getCharacterPositionInLine(),
       };
     }
+    if (this.isEmptyWhereError(error)) {
```
dobladez, line 118: Idea:  adding a new error type means adding another `if` here, and another "predicate" down there. May be you could keep everything about each error close together, and factor-out the common code (like building the returning map).

Roughly, what I'm thinking of would look like this:
```typescript
  knownErrors = [
    {
      type: Soql.ErrorType.EMPTY,
      message: Messages.error_empty,
      predicate: (error: ParserError, context?: ParseTree): boolean =>
        this.parseTree instanceof ParserRuleContext &&
        this.parseTree.start.type === Token.EOF,
    },
    { type: Soql.ErrorType.NOSELECT,
      message: Messages.error_noSelections,
      predicate: (error: ParserError, context?: ParseTree): boolean =>
        context instanceof Parser.SoqlSelectClauseContext &&
        context.exception instanceof InputMismatchException &&
        !hasNonErrorChildren(context)
      )
    }
    // ...etc...

  ];
  public identifyError(error: ParserError): Soql.ModelError {
    const context = this.matchErrorToContext(error);
    const e = this.knownErrors.find(knownError => knownError.predicate(error, context));

    return e? {
      type: e.type,
      message: e.message,
      lineNumber: error.getLineNumber(),
      charInLine: error.getCharacterPositionInLine(),
    } :
     {
      type: Soql.ErrorType.UNKNOWN,
      message: error.getMessage(),
      lineNumber: error.getLineNumber(),
      charInLine: error.getCharacterPositionInLine(),
    };
}
```
So, adding a new error is just adding an element on `knownErrors`;

dobladez (approved):
Looks good. I've suggested a couple of refactorings but they are just that: suggestions.

packages/soql-model/src/serialization/deserializer.test.ts
jgellin-sf: Great suggestion. See `refactor around consistent coding patterns` commit.

packages/soql-model/src/serialization/deserializer.ts
jgellin-sf: Great suggestion. See `refactor around consistent coding patterns` commit.

packages/soql-model/src/messages/messages.ts
```diff
@@ -12,13 +12,20 @@ export namespace Messages {
...
-
+  export const error_emptyWhere = 'Incomplete WHERE clause. The WHERE clause must contain a condition.';
+  export const error_incompleteNestedCondition = 'Incomplete condition. A closing parenthesis is required.';
+  export const error_incompleteAndOrCondition = 'Incomplete condition. Conditions before and after the AND or OR keyword are required.';
+  export const error_incompleteNotCondition = 'Incomplete condition. NOT must be followed by a condition.';
+  export const error_unrecognizedCompareValue = 'Unrecognized comparison value.';
+  export const error_unrecognizedCompareOperator = 'Unrecognized comparison operator.';
+  export const error_unrecognizedCompareField = 'Unrecognized comparison field.';
+  export const error_noCompareValue = 'Incomplete condition. Comparison value is required.';
+  export const error_noCompareOperator = 'Incomplete condition. Comparison operator and value is required.';
```
dehru, line 23: we had a merge conflict here.  i fixed it.  can you verify this looks correct.

dehru (approved):

packages/soql-model/src/messages/messages.ts
dehru, line 23: @ekapner - @jgellin-sf - Will these messages be visible to users?

dehru: Previously, when the existing soql query was opened and had errors, we simply blocked the UI, or placed an asterisk next to a section of the UI.  Hopefully, we can re-use the same message display mechanism we are developing now for symantic errors when the field type doesn't match the comparator or value.

packages/soql-model/src/serialization/deserializer.ts
jonnyhork: That's a pretty cool way to organize the code @dobladez, thanks for sharing the example.

dobladez: Thank you @jonnyhork! Glad you found it useful.

PR #123, opened 2021-01-21
Support IN, NOT IN, INCLUDES, EXCLUDE operators

jgellin-sf:
### What does this PR do?
This PR adds SOQL Builder support for IN, NOT IN, INCLUDES, and EXCLUDES operators. 
- WHERE clauses containing conditions with these operators are no longer unsupported in the SOQL model
- The operators are added to the SOQL Builder UI
- The operators have specific validation, in which the input string is split by comma into separate values, which are then each validated by type
- Recognizable parser errors are identified by the model deserializer for possible recovery

When a multi-value operator is selected, the input must be a comma-separated list of values. The values are not manipulated or normalized as with other operators, so strings must be surrounded by single quotes.

@W-8675683@


packages/soql-model/src/messages/messages.ts
```diff
@@ -21,11 +21,16 @@ export namespace Messages {
...
+
   export const error_fieldInput_boolean = 'Value must be TRUE or FALSE';
   export const error_fieldInput_currency = 'Currency value is not valid';
   export const error_fieldInput_date = 'Date value is not valid';
   export const error_fieldInput_float = 'Value must be numeric';
   export const error_fieldInput_integer = 'Value must be a whole number';
   export const error_fieldInput_picklist = 'Value must be one of: {0}';
+  export const error_fieldInput_string = 'Value must be enclosed in single quotes';
+  export const error_fieldInput_list = 'Input must be a comma separated list of values';
+
```
jgellin-sf, line 34: @ekapner A few new messages

packages/soql-builder-ui/src/modules/querybuilder/services/model.ts
```diff
@@ -54,31 +55,58 @@ export interface ToolingModelJson extends JsonMap {
...
     value: 'GT_EQ',
-    displayValue: '≥'
+    displayValue: '≥',
+    modelValue: Soql.CompareOperator.GT_EQ
   },
   {
     value: 'LIKE',
-    displayValue: 'like'
+    displayValue: 'like',
+    modelValue: undefined
```
dehru, line 89: Why undefined here?

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.test.ts
```diff
@@ -19,15 +19,25 @@ describe('SoqlUtils', () => {
...
-          operator: 'EQ',
-          criteria: { type: 'STRING', value: "'pwt'" },
+          condition: {
+            field: { fieldName: 'Name' },
+            operator: '=',
+            compareValue: {
+              type: 'STRING',
+              value: "'pwt'"
+            }
+          },
```
dehru, line 29: what's the reasoning for this change?

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
```diff
@@ -131,31 +119,62 @@ function convertUiModelToSoqlModel(uiModel: ToolingModelJson): Soql.Query {
...
+      const CONDITION_TYPE_LIKE = 3;
+      let conditionType = CONDITION_TYPE_NONE;
+      if (uiModelCondition.operator) {
+        conditionType = Object.keys(Soql.CompareOperator).map(key => Soql.CompareOperator[key]).includes(uiModelCondition.operator)
+          ? CONDITION_TYPE_FIELDCOMPARE
+          : Object.keys(Soql.InOperator).map(key => Soql.InOperator[key]).includes(uiModelCondition.operator)
+            ? CONDITION_TYPE_IN
+            : Object.keys(Soql.IncludesOperator).map(key => Soql.IncludesOperator[key]).includes(uiModelCondition.operator)
+              ? CONDITION_TYPE_INCLUDES
+              : CONDITION_TYPE_NONE
```
dehru, line 139: head spinning.  i see what you are doing with these ternaries.  I which there was a better way to do this by flipping the data structure so you could just do `Soql.CompareOperator[uiModelCondition.operator]`

packages/soql-builder-ui/src/modules/querybuilder/services/model.ts
jgellin-sf: The parser (and resulting soql-model) is not as tidy as the field-operator-value trifecta of the UI. Simple comparisons (=, !=, <, >, <=, >=) have values that are parsed as literals. LIKE conditions are parsed differently because like values have those `%` wildcards. Similarly the multi-value operators are parsed differently. Also, IN can take an nested inner query as a value. For these reasons we end up with different condition objects for different operators: FieldCompareCondition, LikeCondition, InListCondition, IncludesCondition. FieldCompareCondition has an operator field which can be one of the simple operators. InListCondition can have IN and NOT IN. IncludesCondition can have INCLUDES or EXCLUDES. LikeCondition has only one possible operator (LIKE) so there is no operator field. Hence the 'undefined'. The way it works logically is if the operator is undefined we know it's a LIKE condition. Is this ugly? Yes.

Now. It's not like we are stuck with this. It's our model we can do whatever we want. Maybe we add an operator to the LikeCondition model. It's only ever going to have the LikeOperator.LIKE value, but makes this code a little nicer. What do you think?

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.test.ts
jgellin-sf: Because these multi-value operators don't conform to the previous single field-single operator-single value assumption, we switch from using the ad hoc UI model to the actual soql-model object. This condition object is the same JSON as `new Impl.FieldCompareConditionImpl(new Impl.FieldRefImpl('Name'), Soql.CompareOperator.EQ, new Impl.LiteralImpl(Soql.LiteralType.String, "'pwt'")`.

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
jgellin-sf: Between this and the `undefined` comment above I have some changes in mind for soql-model to make it work for us a little better for us. Stay tuned...

packages/soql-builder-ui/src/modules/querybuilder/services/model.ts
dehru: Yeah, i think it would cause me a little less mental sandpaper to see something there.

packages/soql-model/src/messages/messages.ts
```diff
@@ -21,11 +21,16 @@ export namespace Messages {
   export const error_unrecognizedCompareField = 'Unrecognized comparison field.';
   export const error_noCompareValue = 'Incomplete condition. Comparison value is required.';
   export const error_noCompareOperator = 'Incomplete condition. Comparison operator and value is required.';
+  export const error_incompleteMultiValueList = 'Incomplete values list. Values must be in parentheses and separated by commas.'
```
ekapner, line 24:
```suggestion
  export const error_incompleteMultiValueList = 'Incomplete values list. Place values in parentheses, separated by commas.'
```

packages/soql-model/src/messages/messages.ts
```diff
@@ -21,11 +21,16 @@ export namespace Messages {
...
+
   export const error_fieldInput_boolean = 'Value must be TRUE or FALSE';
   export const error_fieldInput_currency = 'Currency value is not valid';
   export const error_fieldInput_date = 'Date value is not valid';
   export const error_fieldInput_float = 'Value must be numeric';
   export const error_fieldInput_integer = 'Value must be a whole number';
   export const error_fieldInput_picklist = 'Value must be one of: {0}';
+  export const error_fieldInput_string = 'Value must be enclosed in single quotes';
```
ekapner, line 32:
```suggestion
  export const error_fieldInput_string = 'Enclose value in single quotes';
```

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
jonnyhork: I second the head spinning part

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -78,47 +107,71 @@ export default class WhereModifierGroup extends LightningElement {
...
+
+  isMulipleValueOperator(operatorValue: string): boolean {
+    return (operatorValue === 'IN'
+      || operatorValue === 'NOT_IN'
+      || operatorValue === 'INCLUDES'
+      || operatorValue === 'EXCLUDES');
   }
+  // @api get criteria() {
+  //   return this._criteria;
+  // }
```
jonnyhork, line 166: Not sure if this is still WIP but I wanted to check to see if this commented out code should be here?

jonnyhork (commented):
Wow, there is a lot here to digest!

packages/soql-builder-ui/src/modules/querybuilder/services/model.ts
jgellin-sf: I reworked the condition model to make this code a little less squishy. See the `Refactor condition model to simplify code` commit.

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
jgellin-sf: This code is simpler after refactoring the condition model a bit. See the `Refactor condition model to simplify code` commit.

jgellin-sf:
@ekapner Thanks for the review. Your suggestions are committed.

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
```diff
@@ -123,49 +123,47 @@ function convertUiModelToSoqlModel(uiModel: ToolingModelJson): Soql.Query {
...
         ? new Impl.FieldRefImpl(uiModelCondition.field.fieldName)
         : undefined;
 
+      enum ConditionType {
+        FieldCompare = 0,
+        In = 1,
+        Includes = 2
+      }
+      let conditionType = ConditionType.FieldCompare;
+      switch (uiModelCondition.operator) {
```
jonnyhork, line 137: nice, this looks much cleaner!

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
```diff
@@ -131,31 +119,60 @@ function convertUiModelToSoqlModel(uiModel: ToolingModelJson): Soql.Query {
...
+    const simpleGroupArray = uiModel.where.conditions.map(condition => {
+      const uiModelCondition = condition.condition;
+      let returnCondition = undefined;
+
+
+      const field = uiModelCondition.field && uiModelCondition.field.fieldName
+        ? new Impl.FieldRefImpl(uiModelCondition.field.fieldName)
+        : undefined;
+
+      enum ConditionType {
```
dehru, line 131: would it make sense to define these outside of this soqlUtils function?  Is it usable elsewhere?

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -47,6 +47,35 @@ export default class WhereModifierGroup extends LightningElement {
...
 
+  @api get condition(): JsonMap {
+    return this._condition;
+  }
+  set condition(condition: JsonMap) {
+    this._condition = condition;
+    this._criteriaDisplayValue = '';
+    if (this._selectedOperator && this.isMulipleValueOperator(this._selectedOperator.value)) {
+      if (Array.isArray(condition.values)) {
+        this._criteriaDisplayValue = condition.values.map(value => value.value).reduce((soFar, next) => {
```
dehru, line 58: Array.join() is a simpler implementation here that might work.
`condition.values.map(value => value.value).join(', ');`

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
jgellin-sf: Basically it only exists to tell us what kind of model object to create from a non-model object, so it might only apply here.  I don't know where else we'd use it.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -78,46 +107,59 @@ export default class WhereModifierGroup extends LightningElement {
...
   }
 
   get filteredFields() {
     return this.allFields.filter((field) => {
-      return field !== this.selectedField;
+      return field !== this.getFieldName();
     });
   }
 
+  getFieldName(): string | undefined {
```
dehru, line 123: Can you replace this with optional chaining?
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
```diff
@@ -11,9 +11,7 @@ import { JsonMap } from '@salesforce/ts-types';
 import { AndOr } from '../services/model';
 
 interface ConditionTemplate {
-  field: undefined;
-  operator: undefined;
-  criteria: { type: null; value: null };
+  condition: any;
```
dehru, line 14: Can this be a JsonMap instead of any?

dehru:
I think the whereModifierGroup.test.ts file should include a test for the new multipleValueOperator to verify, even though that logic is exported from soqlUtils.ts.  Just something minor and quick.  And additionally, validating the `modifiergroupselection ` event sends the right condition.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
jgellin-sf: Updated

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
jgellin-sf: Webpack doesn't seem to be able to handle optional chaining.

packages/soql-builder-ui/src/modules/querybuilder/where/where.ts
jgellin-sf: Yup. Done

jgellin-sf:
@dehru 
image

See `Add test for multi-value operator` commit

dehru (changes requested):
This could be operator error.  But i'm getting a validation operator error with includes/excludes.  See movie.
https://user-images.githubusercontent.com/599418/106172921-97621e80-6150-11eb-8374-8b2620ad28d2.mov

And an issue converting where in from text editor into builder whereModifierGroup display.


https://user-images.githubusercontent.com/599418/106173005-b06acf80-6150-11eb-8776-a03b0d7cbe49.mov

jgellin-sf:
@dehru Should be fixed.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
dehru: oh yeah, i remember that now.

packages/soql-model/src/model/model.ts
```diff
@@ -195,13 +190,12 @@ export interface Condition extends SoqlModelObject {
...
-  // distance CompareOperator value => UnmodeledSyntax
-  // field LIKE value => LikeCondition
-  // field IncludesOperator ( values ) => IncludesCondition
-  // field InOperator ( semi-join ) => UnmodeledSyntax
-  // field InOperator ( values ) => InListCondition
+  // field ConditionOperator value => FieldCompareCondition
+  // calculation ConditionOperator value => UnmodeledSyntax
+  // distance ConditionOperator value => UnmodeledSyntax
+  // field [Includes|Excludes] ( values ) => IncludesCondition
+  // field [In|NotIn] ( semi-join ) => UnmodeledSyntax
```
dehru, line 197: So is this just a way to document that a field that is a semi-join with In, NotIn syntax, is currently unmodeled?  I remember you showing this before in the walkthrough...but can't remember the specifics

packages/soql-model/src/validators/validatorFactory.test.ts
```diff
@@ -127,4 +127,64 @@ describe('ValidatorFactory should', () => {
...
+      { type: SObjectFieldType.Phone },
+      { type: SObjectFieldType.Picklist },
+      { type: SObjectFieldType.Reference },
+      { type: SObjectFieldType.String },
+      { type: SObjectFieldType.TextArea },
+      { type: SObjectFieldType.Time },
+      { type: SObjectFieldType.Url }
+    ];
+    const expected = [
+      'MultipleInputValidator',
```
dehru, line 160: will these every change to something else based on field type?

dehru (approved):

Issue #120, opened 2021-01-20
With Autosave Off, Preview does not reflect the state of the SOQL Builder UI


dehru:
**OS and version**:
Windows 10

**VS Code Version**:
1.52.1

**SFDX CLI Version**:
7.82.1

### Summary
When autosave is not turned on, the soql preview does not update to represent the actual state of the query.  It appears that the onDocumentChange event does not get fired until the file is saved.  And then it updates.  This can cause confusion since the state of the inputs and the preview do not match.  When the query is run, it is run with the soql query that is currently unsaved in the text document.
_Short summary of what is going on or to provide context_.

### Steps To Reproduce:

1.  Open a soql query in soql builder
1.  Make sure autosave is off in your vscode preferences
1. Adjust the query using the soql-builder form inputs
1. Watch the preview panel of the soql builder is not updated
1. Run the query and watch the results represent the state of the UI ( and the state of the unsaved soql query )
1. Save the file and watch the preview update to reflect the true state of the soql statement.

### Expected result
I guess that the soql statement in the preview should align with the state of the UI controls.  And the state of the unsaved soql file.  It might be nice for the user to know that this preview is not saved.  Which could give them confidence to make changes and manipulate their query until they WANT to save the results.  

I usually work with autosave on.  So I expect the state of the UI, the preview, and the saved file to always be the same.  But novice users may prefer to manually control saving the file.

### Actual result
UI controls in the soql builder and the preview are not in sync.

### Additional information

https://user-images.githubusercontent.com/599418/105215831-06b38f00-5b0f-11eb-9315-bf4624ae1f3f.mov

PR #125, opened 2021-01-21
Fix preview not updating when autosave is off - issue 120

dehru:
### What does this PR do?

This PR address this issue that was introduced in newly published version 50.14.0.  The issue is captured here. https://github.com/forcedotcom/soql-tooling/issues/120

https://github.com/forcedotcom/soql-tooling/issues/120

### DIFF
--- a/packages/soql-builder-ui/package.json
+++ b/packages/soql-builder-ui/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/soql-builder-ui",
   "description": "SOQL Builder UI with LWC",
-  "version": "0.0.28",
+  "version": "0.0.29",
   "author": "Salesforce",
   "publishConfig": {
     "access": "public",
--- a/packages/soql-builder-ui/src/modules/querybuilder/services/toolingModelService.ts
+++ b/packages/soql-builder-ui/src/modules/querybuilder/services/toolingModelService.ts
@@ -28,6 +28,7 @@ export interface ToolingModel extends IMap {
   limit: string;
   errors: List<Map>;
   unsupported: List<Map>;
+  originalSoqlStatement: string;
 }
 // Public inteface for accessing modelService.query
 export interface ToolingModelJson extends JsonMap {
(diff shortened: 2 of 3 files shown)


packages/soql-builder-ui/src/modules/querybuilder/services/toolingModelService.ts
```diff
@@ -28,6 +28,7 @@ export interface ToolingModel extends IMap {
   limit: string;
   errors: List<Map>;
   unsupported: List<Map>;
+  originalSoqlStatement: string;
```
dehru, line 31: The preview component is reactive to this property.  Previously, we were only displaying the soql query from the text document when we got a onDidChangeTextDocument event.  But when the user interacts with the UI, we also generate the soql statement.  So we can use the soql statement generated by the user interaction with the form to also display the soql statement in the preview by adding this property to the immutable model.

packages/soql-builder-ui/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/soql-builder-ui",
   "description": "SOQL Builder UI with LWC",
-  "version": "0.0.28",
+  "version": "0.0.29",
```
dehru, line 4: ready to publish a new version once this is QAd

jonnyhork (approved):
Nice, this looks good. I can't remember why we didn't do it this way before? Also, do we know the source of the bug, when and what changed?

dehru:
> Nice, this looks good. I can't remember why we didn't do it this way before? Also, do we know the source of the bug, when and what changed?

So we used to always show the preview from the soql statement that was generated by the UI.  But then we had that issue where the user could open a soql query from a text editor that we didn't support...but we still wanted to show them the soql statement and allow them to run what they see.  So we switched it to only update the preview from the incoming soql statement.

I think what we want is show the soql statement from the text editor when the change event happens.  But also, when the user interacts with our form, immediately update the soql statement.  Since we are sending that to the text editor anyway.

It's like the best of both worlds now.  I hope.

jonnyhork:
QA Passed for me

PR #126, opened 2021-01-22
Add smarter completion on GROUP BY clause and WHERE literals

dobladez:
### What does this PR do?

Add smarter completion on GROUP BY clause.

Also:
 * Complete only with "groupable" fields.
 * Give priority to complete GROUP BY with fields which appear
   on SELECT and are not aggregated or grouped.
 * On ORDER BY: Complete only with "sortable" fields
 * Add completion inside semi-join SELECTs (in WHERE clauses)
 * Fix defect when cursor is around multiple newlines
 * Improve query analysis impl
 * Pre-select "WHERE" option if provided
 * Refactor: move soql operator semantics to LSP server-side
 * Add more code-completion proposals for literal values, including range date literals (LAST_YEAR, YESTERDAY, LAST_N_DAYS:n, etc.) and simple templates for: int, double, string, currency

To test it on the VS Code extension, use branch `feature/soql-code-completion-GROUPBY` of `salesforcedx-vscode`.


packages/language-server/src/completion.ts
```diff
@@ -447,30 +564,115 @@ function withSoqlContext(
...
+        ...soqlContext,
+        ...{ onlyTypes: ['date', 'datetime'] },
+      })
+    ),
+    ...soqlParametricDateRangeLiterals.map((k) =>
+      withSoqlContext(newSnippetItem(k, k.replace(':n', ':${1:nn}') + '$0'), {
+        ...soqlContext,
+        ...{ onlyTypes: ['date', 'datetime'] },
+      })
+    ),
```
jgellin-sf, line 659: Nice

packages/language-server/src/completion.ts
```diff
@@ -410,34 +500,61 @@ function isCursorAfter(
...
-  notNillable?: boolean;
+  onlyTypes?: string[];
+  onlyAggregatable?: boolean;
+  onlyGroupable?: boolean;
+  onlySortable?: boolean;
+  onlyNillable?: boolean;
+  mostLikelyItems?: string[];
 }
 
 function withSoqlContext(
```
jgellin-sf, line 560: This context pattern turned out really useful.⭐️⭐️⭐️

jgellin-sf (dismissed):
It looks like eventually we will have SObject information on the language server, but I really like the choices you've made to organize the semantics code in the absence of that information. (i.e. Collect every possible completion candidate, filter based on context later.)

packages/language-server/src/completion.ts
```diff
@@ -212,28 +250,38 @@ function generateCandidatesFromTokens(
...
+        const soqlOperator = soqlOperators[itemText];
+        soqlItemContext.onlyTypes = soqlOperator.types;
+      }
+    }
+
     // Some "manual" improvements for some keywords:
     if (['IN', 'NOT IN'].includes(itemText)) {
       itemText = itemText + ' (';
     } else if (['INCLUDES', 'EXCLUDES', 'DISTANCE'].includes(itemText)) {
       itemText = itemText + '(';
```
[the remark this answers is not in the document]
dobladez: Good catch!  The intention was to use the space only on IN, NOT IN, INCLUDES and EXCLUDES (like the examples on the documentation), I just fixed it, thanks!
I think the rationale is that these are operators (which expect a list), while DISTANCE(...), AVG(...), MAX(...), etc. are function applications.

PR #129, opened 2021-01-27
[WAIT] Dehru/where error messages recoverable

dehru:
### What does this PR do?
This PR expands on the ability of soql builder to display errors and still let the user interact with the builder ui.  Unsupported messages are still blocked, but the user can override that block and interact with the ui if they really want to.

Emily is going to need to help us with the correct messaging to the user.

@W-8777407@

There are some messaging work related to really smooth out this ui.  those stories are:

- W-8794016
- W-8793884
- W-8793969

https://user-images.githubusercontent.com/599418/105924673-011ce400-5ffc-11eb-9567-1c5f96bb91e9.mov


packages/soql-builder-ui/src/modules/querybuilder/app/app.html
```diff
@@ -12,31 +12,36 @@
...
-                Your SOQL statement contains syntax errors. Use a text editor to
-                edit the query, and try again.
-              </p>
+            <div class="block-message warning-notification">
+              <h3>Unsupported Syntax</h3>
+              Your query contains statements that SOQL Builder doesn't currently
+              support. You should use a text editor to edit your query. If you
+              edit the query using SOQL Builder, any unsupported syntax will be
+              overwritten.
+              <ul>
```
dehru, line 30: @ekapner 
Screen Shot 2021-01-27 at 9 39 29 AM

packages/soql-builder-ui/src/modules/querybuilder/header/header.html
```diff
@@ -7,14 +7,25 @@
...
-      </template>
-      <template if:true={isRunning}>
-        <button class="btn--disabled">Running...</button>
-      </template>
+  <header class={headerClasses}>
+    <template if:true={hasErrors}>
+      <div class="header__content">
+        <h3>Syntax Error</h3>
+        Your query contains errors that SOQL Builder cannot parse.
+        <a class="show-more" onclick={handleShowErrors}>({triggerText})</a>
```
dehru, line 15: Screen Shot 2021-01-27 at 9 42 24 AM

User can see the errors by clicking ( show errors )
Screen Shot 2021-01-27 at 9 42 37 AM

dehru:
Also, note that it is possible to have both the Syntax Error and the Unsupported Syntax at the same time.

Screen Shot 2021-01-27 at 9 43 08 AM

packages/soql-builder-ui/src/modules/querybuilder/app/app.html
```diff
@@ -12,31 +12,36 @@
...
-              <p>
-                Your SOQL statement contains syntax errors. Use a text editor to
-                edit the query, and try again.
-              </p>
+            <div class="block-message warning-notification">
+              <h3>Unsupported Syntax</h3>
+              Your query contains statements that SOQL Builder doesn't currently
+              support. You should use a text editor to edit your query. If you
+              edit the query using SOQL Builder, any unsupported syntax will be
+              overwritten.
```
ekapner, line 29: Your query contains statements that SOQL Builder doesn't currently
              support. You can use a a text editor to edit the query. If you
             select Edit the Query Anyway, SOQL Builder removes the unsupported syntax.

ekapner: @dehru - what do you think?

packages/soql-builder-ui/src/modules/querybuilder/header/header.html
```diff
@@ -7,14 +7,25 @@
...
-        <button class="run-button" onclick={handleRunQuery}>Run Query</button>
-      </template>
-      <template if:true={isRunning}>
-        <button class="btn--disabled">Running...</button>
-      </template>
+  <header class={headerClasses}>
+    <template if:true={hasErrors}>
+      <div class="header__content">
+        <h3>Syntax Error</h3>
+        Your query contains errors that SOQL Builder cannot parse.
```
ekapner, line 14:
```suggestion
        Your query contains errors that SOQL Builder can't parse.
```

ekapner (commented):
Left some edits like we discussed. Did I review everything?

packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
```diff
@@ -214,3 +248,13 @@ button:active {
...
   cursor: pointer;
 }
+
+.show-more {
+  cursor: pointer;
+  text-decoration: underline;
+  margin-left: 0.25rem;
+}
+
+.center {
```
jonnyhork, line 258: Maybe this should be `text--center`, as the class name `center` is misleading. It will not center anything _but_ text.

packages/soql-builder-ui/src/modules/querybuilder/header/header.test.ts
```diff
@@ -21,17 +31,79 @@ describe('Header', () => {
     }
   });
 
-  it('emits a run event', () => {
+  it('emits a run event', async () => {
```
jonnyhork, line 34: I haven't found `async` to be necessary when I use the `return Promise.resolve()...`syntax for testing. I don't get any warnings from Jest and the test runs fine. I wonder if its because we are calling `resolve` ourself so there is nothing to await.

looks like its only needed if we use [async/await](https://jestjs.io/docs/en/tutorial-async#asyncawait)

packages/soql-builder-ui/src/modules/querybuilder/header/header.ts
```diff
@@ -9,9 +9,40 @@
...
 export default class Header extends LightningElement {
   @api isRunning = false;
+  @api errorMessages = [];
+  get hasErrors() {
+    return this.errorMessages.length > 0;
+  }
+  get headerClasses() {
+    return this.hasErrors ? 'warning-notification' : '';
+  }
+  get triggerText() {
```
jonnyhork, line 19: // i18n

packages/soql-builder-ui/src/modules/querybuilder/header/header.ts
```diff
@@ -9,9 +9,40 @@
...
+
+  get isButtonDisabled() {
+    return this.hasErrors || this.isRunning;
+  }
+
+  get buttonClasses() {
+    return this.isButtonDisabled ? 'btn--disabled' : '';
+  }
+
+  get buttonText() {
```
jonnyhork, line 32: // i18n

jonnyhork (dismissed):
This is a huge improvement from what we had, looks good to me!

jgellin-sf:
The code looks good. 

We have an interesting problem on our hands concerning unsupported syntax behavior. I like that fact that we are allowing users to edit anyway. That's the most important thing. I think there's a balance to be reached with whether we overwrite. If the unsupported stuff is in the part of the language that can be expressed with our editor, then ok, we need to overwrite, but if it's outside what can be expressed (GROUP BY, WITH, etc.) maybe we should try and maintain that. Anyway, something to revisit in another PR.

packages/soql-builder-ui/src/modules/querybuilder/cssCommon/cssCommon.css
dehru: done.

packages/soql-builder-ui/src/modules/querybuilder/header/header.test.ts
dehru: yeah, it does look like if we return a promise, jest handles it.  I'm always afraid of a false positive, since that's what used to happen to me.  

https://jestjs.io/docs/en/asynchronous#promises

PR #130, opened 2021-01-27
Jhork/custom select from

jonnyhork:
### What does this PR do?
This PR will add support a single selection mode of the Custom Select component and other improvements. This PR will also replace the standard `<select>` element with the Custom Select component to search and chose Objects in the SOQL builder.

**New Features**
- `@api value : string[]` can be used to query state of the component for selections made independently of the model. This will help with the WHERE clause when we need to know that a selection has been made but wait until all modifiers have a value before updating the model.
- If there are multiple Custom Selects on the page then only one will be allowed to display its options at a time.
- The width of the component will be determined by its parent element up to a max width. NOTE: currently the options menu will still span the full width of the form column.
- Added a arrow icon `⌃` that will match standard `<select>`elements and toggle the options menu open or closed.
- The placeholder text will fade out when the user clicks on the input
- Added comments for posterity
 
 **Single Select Mode**
- If there is a selected option, then it will be displayed as the value of the component.
- When the user clicks on the input the value will be highlighted (selected) so the user can start typing to replace it.

https://user-images.githubusercontent.com/25188632/106027995-a7163000-6088-11eb-9a61-a2f811c094ef.mov

@W-8616118@


packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
```diff
@@ -149,8 +217,100 @@ describe('Custom Select', () => {
...
+    describe('Single Select Mode', () => {
+      beforeEach(() => {
+        customSelect.multiple = false;
+      });
+      /*
+      During testing the getter for the displayValue() is called once
+      & before the renderedCallback that caches the inputSelectEl.
+      The only way I have been able to trigger another render of the component
+      is to click the search bar.
+       */
```
jonnyhork, line 230: I might want to post a question to the LWC framework channel asking about where the getters fit into the lifecycle of the component. I'm am surprised to see the getter called before the rendered callback and that an element is not in the DOM when we query for it in the `renderedCallback()`

jonnyhork: For now I can do a click event to get everything in place for the test but its not ideal.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
```diff
@@ -149,8 +217,100 @@ describe('Custom Select', () => {
...
+        document.body.appendChild(customSelect);
+        expect(customSelect.multiple).toBe(true);
+        const searchBar = getInputSearchBar();
+        expect(searchBar.value).toBe('');
+        expect(customSelect.value).toBe(customSelect.selectedOptions);
+      });
+    });
+
+    describe('Dropdown Arrow', () => {
+      it('should toggle the options menu when clicked, with directional arrow', () => {
```
jonnyhork, line 286: I thought about breaking this up into 3 tests but found I repeating 90% of the code so I just combined it.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -32,23 +51,80 @@ export default class CustomSelect extends LightningElement {
...
+    }
+
+    if (this.isSingleSelect && this.selectInputEl) {
+      if (!this.selectInputIsFocused || this.selectInputEl.value.length) {
+        return this._value[0] || '';
+      }
+    }
+
+    return '';
+  }
```
jonnyhork, line 75: This method is used to control what displayed in the search bar as the value. I need to control the display logic and the state of the component separately.

I found that getting the behavior I wanted was tricky, and this might be something that is just inherently not intuitive. Without `|| this.selectInputEl.value.length` in line 69, when the user would delete the last character of the input then the selected value would immediately repopulate in the input instead of showing the placeholder.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -95,6 +179,13 @@ export default class CustomSelect extends LightningElement {
           value: optionValue
         }
       });
+
+      if (this.isSingleSelect) {
+        this._value = [optionValue];
+      } else {
+        this._value = [...this._value, optionValue];
+      }
```
jonnyhork, line 187: This might need to change when we want to implement true multi selection but the intention is to maintain state in the component with having to read/write to the model.

packages/soql-builder-ui/src/modules/querybuilder/from/from.ts
```diff
@@ -10,27 +10,23 @@ import { LightningElement, api } from 'lwc';
...
+  @api
+  get selected() {
+    return this._selectedObject[0];
   }
 
-  get defaultOptionText() {
-    // TODO: i18n
-    return this.isLoading ? 'Loading...' : 'Select object...';
+  set selected(objectName: string) {
+    this._selectedObject = objectName ? [objectName] : [];
```
jonnyhork, line 24: The custom select component accepts an array of selections so I needed to re package the values from the model.

packages/soql-data-view/web/queryDataView.css
```diff
@@ -17,8 +17,8 @@ body {
   --soql-main-min-width: 480px;
   --soql-main-max-width: 960px;
   --soql-media-max-width: 480px;
-  --soql-label-width: 80px;
-  --soql-input-width: 400px;
+  --soql-label-container-width: 80px;
+  --soql-selection-container-width: 400px;
```
jonnyhork, line 21: I changed this now so that the dataView and the builder UI are in sync.

jonnyhork (commented):
Some of the behavior described in the PR comments may be hard to envision so if anyone wants videos I can make them.

I think a lot of this work will set us up to use this component for ODER BY and WHERE, hopefully with only changes needed in the parent components from here until we had multi -select support

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
dehru: I wonder if the gitter is called during the render process.  And then when it's finished rendering, we get the renderedCallback()

jonnyhork: I could see that. Does that mean that we are not using the right pattern to capture or access elements in the DOM between the getter and rendered callback? Something to consider.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -32,23 +53,80 @@ export default class CustomSelect extends LightningElement {
...
+    classList += this.optionListIsHidden ? '' : ' select__dropdown-arrow--up';
+    return classList;
+  }
+
+  get isSingleSelect() {
+    return !this.multiple;
+  }
+
+  get isMultipleSelect() {
+    return this.multiple;
```
jonnyhork, line 98: I realized this api as it is might be a bit odd.
`this.multiple` is set as boolean value `false` in the component. However, because LWC won't let us use literals in the template `multiple={true}` we have to pass in a string `multiple="true"`. 
Unfortunately, as it stands, that means that I could put anything as the the string value and it would be coerssed as boolean `true` so `multiple = ''false"` passed in from parent to child is also true.

To fix this I would have to set the default value to string "false" and only work with string values "true" and "false"
```javascript
get isSingleSelect() {
    return this.multiple === "false" || this.multiple === undefined;
  }
  ```
  Or it would depend on the parent to use a getter to pass a boolean value to the child
  ```javascript
  // in the parent
  get isMultiple() {
    return true;
  }
  // html
  multiple={isMultiple}
  ```
  
I do not like the option with a getter in the parent. Sicme we are the only ones using this component at the moment I can see leaving it as is for now. I also don't mind adding logic to check explicitly for strings "true" or "false" but it might be a but verbose for a simple check, idk.
Thoughts @dehru?

File: packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css [hunk 8]
```diff
@@ -45,17 +58,30 @@
   cursor: pointer;
 }
 
+.select__dropdown-arrow {
+  transform: rotate(180deg);
+  margin-right: 2px;
```
dehru, line 63: for some reason, I feel the arrow is uncomfortably close to the right.  Just a tad.  Not a change request, just stating a strange gut reaction.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.html
```diff
@@ -1,39 +1,55 @@
...
+          key={optionName}
+          class="option"
+          data-option-value={optionName}
+        >
+          {optionName}
+        </p>
+      </template>
+      <template if:true={noResultsFound}>
+        <!-- i18n -->
+        <p class="option option--placeholder">No results found.</p>
```
dehru, line 51: placeholder is a loaded term.  how 'bout option--empty or option--noresults

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
```diff
@@ -80,6 +82,20 @@ describe('Custom Select', () => {
...
+    });
+
+    it('should style the placeholder text when input has focus', () => {
+      document.body.appendChild(customSelect);
+      customSelect.placeholderText = 'hold my beer';
+      let searchBar = getInputSearchBar();
+      searchBar.focus();
+      return Promise.resolve().then(() => {
+        const classList = Array.from(searchBar.classList);
+        expect(classList).toContain('select__input-placeholder--fadeout');
```
dehru, line 98: It's weird to have css classname changes to cause failing tests.  I had that case in my work this week.  it's a valid test...not much we can do about it.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
```diff
@@ -100,6 +116,58 @@ describe('Custom Select', () => {
...
+      document.body.appendChild(customSelect);
+      const searchBar = getInputSearchBar();
+      const optionsWrapper = getOptionsWrapper();
+      let optionsWrapperClassList;
+      let optionsOpenEvent;
+      searchBar.click();
+      return Promise.resolve()
+        .then(() => {
+          optionsWrapperClassList = Array.from(optionsWrapper.classList);
+          expect(optionsWrapperClassList).toContain('options--open');
```
dehru, line 141: might think about making this a const since it's used thrice in this test case.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
```diff
@@ -100,6 +116,58 @@ describe('Custom Select', () => {
...
+              target: newInput
+            },
+            bubbles: true,
+            composed: true
+          });
+          customSelect.dispatchEvent(optionsOpenEvent);
+        })
+        .then(() => {
+          optionsWrapperClassList = Array.from(optionsWrapper.classList);
+          expect(optionsWrapperClassList).not.toContain('options--open');
```
dehru, line 167: very nice

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
dehru: I don't know if there's other options.  besides we introduce an isRendered boolean and then check for that everywhere.  that could also be messy.  i think it's just something we have to deal with.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
jonnyhork: ya know I had it at 5px originally but I noticed it was further away from the standard <select> element so I changed it to match. I also feel like its a little close but the other thought it that we may want more room once the component is smaller in the WHERE form.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.html
jonnyhork: I went with `option--disabled` to be more generalizable. That way if we use it for something else we don't have to change the class again.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.test.ts
jonnyhork: Done. Thanks

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -1,26 +1,61 @@
...
+ *  @attr all-options: string[] - This the list of all possible options
+ *                                the user can select from.
+ *  @attr selected-options: string[] - If present in single-select, the selectedOption
+ *                                     will be displayed as the value of the input.
+ *                                     The list of options rendered will be all-options - selected-options
+ *  @attr placeholder-text: string - This is the value to be displayed as a placeholder for the input.
+ *  @property value: string[] - Will return the currently selected value(s).
+ *  @event option__selection - This is emitted everytime a valid option is selected.
+ *                             detail { value: optionValue }
+ * **/
```
jonnyhork, line 16: I added this to help future dev know how to use this component.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
dehru: Meh, it's currently just us using it.  and messing that up would be just us using it wrong.

dehru:
This is so good i gotta get it merged asap.

PR #134, opened 2021-01-28
Jhork/custom select order by

jonnyhork:
### What does this PR do?
This PR will use the `custom-select` component in the ORDER BY form.

@W-8616120@


dehru:
I think the value of order-by field needs to be cleared when the full order by statement is added.  Right now it retains the value after the order by statement is added.  This should probably be controlled by the order-by component.

https://user-images.githubusercontent.com/599418/106201500-cc349c80-6175-11eb-9d51-44f3e8878b7a.mov

dehru (changes requested):
i attached a movie that shows how i think the state should be cleared

dehru:
OK.  So i played with the implementation a bit and got it to reset back to 'Search fields' instead of holding state.  It was a little big hacky.  But after that, i'm not sure we should change the current behavior.  I think it's fine as is.

dehru (approved):

PR #147, opened 2021-02-01
Unexpected EOF error

jgellin-sf:
### What does this PR do?
This PR creates a catch-all error for any parse error that occurs because of an unexpected end of file. Some of the EOF cases are already caught in known errors such as some of the various 'Incomplete' errors. However, there are some specific cases where EOF errors are not caught, such as an incomplete string token or an incomplete clause that is currently not modeled for the SOQL builder. Since this is a catch-all error, the logic to recognize it can match other EOF errors, so the error identifier must call this recognition predicate last. This is called out as a comment in the code.
```
// incomplete string token (no closing ')
SELECT field FROM object WHERE field = '
// incomplete unmodeled clause (GROUP BY not yet modeled for SOQL builder)
SELECT field FROM object GROUP BY
```
Previously, these cases resulted in an unknown error, and the original unfriendly error message generated by the parser was shown. The recognized error message is: 'Unexpected end of file.'

@W-8794016@


packages/soql-model/src/serialization/deserializer.ts
```diff
@@ -263,7 +270,12 @@ class ErrorIdentifier {
 
   public identifyError(error: ParserError): Soql.ModelError {
     const context = this.matchErrorToContext(error);
-    const knownErrorMatch = this.knownErrors.find(knownError => knownError.predicate(error, context));
+    let knownErrorMatch = this.knownErrors.find(knownError => knownError.predicate(error, context));
+
+    // unexpectedEOF is an EOF catch-all, make sure it is tested last
+    if (!knownErrorMatch && this.unexpectedEOF.predicate(error, context)) {
+      knownErrorMatch = this.unexpectedEOF;
+    }
```
dobladez, line 278: Isn't it enough to keep it at the end of the `knownErrors` array ?

jgellin-sf: Yes but then we need to remember to do that when more known errors are added. Even with a comment I don't trust all future developers (especially myself) to read the comment and keep that error in the correct position. I thought this more explicit callout was more future-proof.

dobladez: I'd suggest we add tests to cover the case of having EOF errors together with other errors.

With those tests in place, I don't think we need to treat `unexpectedEOF` as a special case outside of the `knownErrors` list.   I'd still leave that same comment on the last element of the array.

jgellin-sf: Ok, you've convinced me. See the `elegance` commit.

dobladez: :-)  ... I don't think we have a tests to cover that situation, Do we?  I mean, to assert that "unexpectedEOF" never fires before another known error.

jgellin-sf: I'm struggling with a way to come up with that test without exporting more from the deserializer module (currently only exporting ModelDeserializer, everything else is helper classes). Still thinking about it, but maybe you know some tricks I don't. Any suggestions?

dobladez: Oh, just pass a SOQL text that you expect to cause both an EOF and another error.

Actually, it looks like it's already covered! So we are good 👍:  I just checked out this branch, moved the error up on the `knownErrors` array and with that change, 10 tests fail.

jgellin-sf: Ok good I thought that might cover things. I was trying to come up with an explicit test.

dehru (approved):

dehru:
Before we merge this, let's figure out what Matt wants to do to display these.

PR #149, opened 2021-02-01
Consistent unsupported conditions

jgellin-sf:
### What does this PR do?
This PR refactors the model code slightly to make any condition that cannot currently be supported in the UI an unmodeled syntax object. Previously, non-simple group conditions were marked as "unmodeled:complex-group" by UI code, and later processing was complicated by inconsistent structure. This PR moves all classification of unmodeled syntax to the model code.

@W-8793884@


jonnyhork (approved):
These changes seem simple enough to me and you know this code best. I approve but based on my less than confident understanding of the code flow in the model, someone else should also review it.

dehru:
Before we merge this, let's figure out what Matt wants to do to display these.

dehru:
This has passed QA.  I indeed get a nice message in the new format from unsupported.reason.message.

dehru (approved):
This passes QA for me.  I can see the message instead of the code in the UI.

Screen Shot 2021-02-08 at 3 30 36 PM

PR #150, opened 2021-02-01
set the error overlay to be on top of all other form elements

jonnyhork:
### What does this PR do?
This PR will refactor the use of `x-index` for some UI components. It will also move the error overlay to the further down the DOM of the form section so that it sits on top of other siblings.

### DIFF
--- a/packages/soql-builder-ui/src/modules/querybuilder/app/app.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/app/app.css
@@ -58,7 +58,7 @@ main {
   position: absolute;
   top: 0;
   left: 0;
-  z-index: 201; /* one above the custom select options pane */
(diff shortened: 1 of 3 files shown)


packages/soql-builder-ui/src/modules/querybuilder/app/app.css
```diff
@@ -58,7 +58,7 @@ main {
   position: absolute;
   top: 0;
   left: 0;
-  z-index: 201; /* one above the custom select options pane */
+  z-index: 2; /* one above the custom select options pane */
```
dehru, line 61: did you find you needed to set this and the z-index in custom-select so that the block-message would sit above?  I wondered if maybe dom order was enough

dehru (approved):

PR #152, opened 2021-02-01
Merge Custom Select Components into Main

dehru:
### What does this PR do?

Updates main with the latest from develop branch

Continued LSP improvements
Custom Select Component in Soql Builder UI

### DIFF
--- a/packages/soql-data-view/web/queryDataView.css
+++ b/packages/soql-data-view/web/queryDataView.css
@@ -17,8 +17,8 @@ body {
   --soql-main-min-width: 480px;
   --soql-main-max-width: 960px;
   --soql-media-max-width: 480px;
-  --soql-label-width: 80px;
-  --soql-input-width: 400px;
+  --soql-label-container-width: 80px;
+  --soql-selection-container-width: 400px;
   /* colors */
   --soql-color-light-grey: #ecebea;
   --soql-color-medium-grey: #b0adab;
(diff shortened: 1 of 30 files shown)


packages/soql-data-view/web/queryDataView.css
```diff
@@ -17,8 +17,8 @@ body {
   --soql-main-min-width: 480px;
   --soql-main-max-width: 960px;
   --soql-media-max-width: 480px;
-  --soql-label-width: 80px;
-  --soql-input-width: 400px;
+  --soql-label-container-width: 80px;
+  --soql-selection-container-width: 400px;
```
jgellin-sf, line 21: These properties are never used in the data view. I assume they are changed here for consistency with SOQL builder. But should they be removed instead?

jonnyhork: You are correct, they are not used (yet?). I put them there sort of as documentation so we stay in sync with the builder if we do use them. Kind of like the `unmodeledsyntax`. I could go either way with these, it not that hard to add them back, I just didn't want anyone to forget we have global variables on the builder side.

jonnyhork: I am guessing there is a good chance I will be the one to do any work on the data view anyway, so we can remove them now if we feel that's better.

jgellin-sf: Let's leave them for now.

jgellin-sf (approved):

PR #153, opened 2021-02-01
Jhork/custom select where

jonnyhork:
### What does this PR do?
This PR will use the <custom-select> for selecting fields in the `whereModifierGroup` component.
@W-8616121@

### DIFF
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
@@ -38,7 +38,7 @@ main.custom-select {
 
 .select__input {
   width: 90%;
-  padding: 4px;
+  height: 1.25em;
   margin-left: 24px;
   border: none;
 }
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
@@ -38,7 +38,6 @@ main.custom-select {
 
 .select__input {
   width: 90%;
-  height: 1em;
   padding: 4px;
   margin-left: 24px;
   border: none;
--- a/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.css
@@ -25,7 +25,6 @@ prevent select el from
 expanding to fit content
  */
 .modifier__fields {
-  width: 100%;
   flex-grow: 2;
 }
 
--- a/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.css
@@ -19,22 +19,6 @@
 .modifier__item {
   margin-right: 4px;
 }
-/*
-setting a width will
-prevent select el from
-expanding to fit content
- */
-.modifier__fields {
-  flex-grow: 2;
-}
-
-.modifier__operator {
-  flex-grow: 1;
-}
-
-.modifier__criteria {
-  flex-grow: 1;
-}
 
 .modifier__operator.tooltip--error select,
 .modifier__criteria.tooltip--error input {
--- a/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.css
@@ -34,7 +34,6 @@ expanding to fit content
 }
 
 .modifier__criteria {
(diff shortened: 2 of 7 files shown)


packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -105,7 +126,7 @@ export default class WhereModifierGroup extends LightningElement {
   }
 
   get _selectedField() {
-    return this.getFieldName();
+    return this.getFieldName() ? [this.getFieldName()] : [];
```
jonnyhork, line 129: This property is only used by the custom select component, which expects an array, so I changed the return value here instead of creating a new method just for that purpose.

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.css
```diff
@@ -38,7 +38,6 @@ main.custom-select {
 
 .select__input {
   width: 90%;
-  height: 1em;
```
jonnyhork, line 41: I noticed the bottom of a 'g' was cut off in the input. So I removed this rule and let the height be set in commonCSS.css with the rest of the `input` elements at `1.5em`. That fixed the issue for me.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.css
```diff
@@ -19,24 +19,6 @@
...
-}
-
-.modifier__operator {
-  flex-grow: 1;
-}
-
-.modifier__criteria {
-  padding-left: 4px;
-  flex-grow: 1;
-}
```
jonnyhork, line 39: These rules were not doing what I initially thought they were doing and are no longer needed.

jonnyhork: The padding-left is extra as well and not needed.

jonnyhork (commented):
Most of the changes are formatting changes. We should look into that one day and make sure we all have the same formatting rules.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.html
```diff
@@ -1,24 +1,14 @@
...
-      <template if:false={hasSelectedField}>
-        <option value="" class="placeholder" data-el-default-option>
-          {defaultFieldOptionText}
-        </option>
-      </template>
-      <template for:each={filteredFields} for:item="field">
-        <option key={field} value={field}>{field}</option>
-      </template>
-    </select>
+    <div class="modifier__item modifier__fields">
```
dehru, line 4: is the class modifier__fields used anymore?   It looks like you removed the css class definition.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -320,7 +353,7 @@ export default class WhereModifierGroup extends LightningElement {
       };
       if (isMultiInput) {
         const rawValues = splitMultiInputValues(normalizedInput);
-        const values = rawValues.map(value => {
+        const values = rawValues.map((value) => {
```
dehru, line 356: I wonder why in your PR there are several of these reformatted to wrap the single argument with parens.  Theoretically, this shouldn't be necessary.  

I also noticed in my merge this morning that we've got an eslint config issue. I don't quite understand how eslint and prettier are playing together if at all.  But wondered if that's affecting it.

https://github.com/forcedotcom/soql-tooling/blob/jhork/custom-select-where/packages/soql-builder-ui/.eslintrc.json#L4

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -40,6 +44,7 @@ export default class WhereModifierGroup extends LightningElement {
   criteriaErrorMessage = '';
   hasOperatorError = false;
   hasCriteriaError = false;
+  selectPlaceHolderText = 'Search Fields...';
```
dehru, line 47: // i18n

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.html
jonnyhork: 👍 Fixed in the `remove class, add i18n` commit

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
jonnyhork: 👍 Fixed in the `remove class, add i18n` commit

jonnyhork:
New PR at #161

PR #155, opened 2021-02-02
Human readable messages for unsupported syntax

jgellin-sf:
### What does this PR do?
This PR creates human readable messages for unsupported syntax. The existing 'reason codes', formatted like `unmodeled:<soql element>`, are retained for telemetry purposes.

@W-8793969@

### DIFF
--- a/packages/soql-model/src/messages/messages.ts
+++ b/packages/soql-model/src/messages/messages.ts
@@ -33,4 +33,27 @@ export namespace Messages {
   export const error_fieldInput_list = 'Input must be a comma separated list of values';
 
   export const error_operatorInput = "{0} operator can't be used for this field type";
+
+  export const unmodeled_as = 'Object alias';
+  export const unmodeled_using = 'USING SCOPE clause';
+  export const unmodeled_alias = 'Field alias';
+  export const unmodeled_semijoin = 'Subquery';
+  export const unmodeled_typeof = 'TYPEOF clause';
+  export const unmodeled_distance = 'DISTANCE expression';
+  export const unmodeled_select = 'Unsupported SELECT expression';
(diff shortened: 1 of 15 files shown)


packages/soql-model/src/messages/messages.ts
```diff
@@ -33,4 +33,27 @@ export namespace Messages {
...
+  export const unmodeled_bind = 'BIND clause';
+  export const unmodeled_recordtracking = 'Record tracking clause';
+  export const unmodeled_update = 'Update statistics clause';
+  export const unmodeled_functionreference = 'Function expression';
+  export const unmodeled_colonexpression = 'Colon expression';
+  export const unmodeled_emptycondition = 'Empty condition';
+  export const unmodeled_calculatedcondition = 'Calculated condition field';
+  export const unmodeled_distancecondition = 'DISTANCE condition';
+  export const unmodeled_incolonexpressioncondition = 'Colon expression as IN value';
+  export const unmodeled_insemijoincondition = 'Subquery as IN value';
```
jgellin-sf, line 58: @ekapner In @dehru 's [PR](https://github.com/forcedotcom/soql-tooling/pull/129) there is an implementation for a banner about unsupported syntax which lists items from the given query that are unsupported. These strings are intended to be used as the items in that list. So they are simply descriptions of SOQL language elements. SOQL keywords are all caps.

dehru (approved):

packages/soql-model/src/messages/messages.ts
ekapner: Thanks for the explanation. They look good to me. We can always address if we find any of these items are unclear/cause confusion.

ekapner (approved):

PR #161, opened 2021-02-03
Jhork/where custom select

jonnyhork:
### What does this PR do?
Use Custom Select in the where component
@W-8616121@

### DIFF
--- a/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
+++ b/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
@@ -44,7 +44,7 @@ export default class WhereModifierGroup extends LightningElement {
   criteriaErrorMessage = '';
   hasOperatorError = false;
   hasCriteriaError = false;
-  selectPlaceHolderText = 'Search Fields...';
+  selectPlaceHolderText = 'Search Fields...'; //i18n
 
   handleSelectionEvent: () => void;
   // this need to be public so parent can read value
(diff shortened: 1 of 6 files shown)


packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -320,7 +353,7 @@ export default class WhereModifierGroup extends LightningElement {
       };
       if (isMultiInput) {
         const rawValues = splitMultiInputValues(normalizedInput);
-        const values = rawValues.map(value => {
+        const values = rawValues.map((value) => {
```
dehru, line 356: I'm still wondering why your PR has added parens around these single argument values.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -150,10 +177,12 @@ export default class WhereModifierGroup extends LightningElement {
   }
 
   isMulipleValueOperator(operatorValue: string): boolean {
-    return (operatorValue === 'IN'
-      || operatorValue === 'NOT_IN'
-      || operatorValue === 'INCLUDES'
-      || operatorValue === 'EXCLUDES');
+    return (
```
dehru, line 180: linter / prettier differences?  Not the end of the world but seems like we keep having these type of diffs.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
jonnyhork: yeah we need to compare prettier/linter rules

dehru (approved):

PR #162, opened 2021-02-03
Fix: 'undefined' in SOQL string

dobladez:
### What does this PR do?
Fix spurious "undefined" string appearing when on SOQL query generation with no comments

W-8611529


jgellin-sf (approved):
The truthy check in soqlUtils.ts is probably enough to fix the immediate issue, but the additional checks should protect against future problems.

dehru (approved):
This passes Qa for me

PR #163, opened 2021-02-03
Jhork/workflow updates

jonnyhork:
### What does this PR do?
This PR will update some of the workspace settings to help keep formatting consistant between team members. It will also add a command to `yarn clean` that will clear out the cache in Jest to help avoid errors when testing.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -60,5 +60,7 @@
     "activityBar.background": "#523d89",
     "titleBar.activeBackground": "#523d89",
     "titleBar.activeForeground": "#f3f1ef"
-  }
+  },
+  "prettier.configPath": "../.prettierrc",
+  "editor.defaultFormatter": "esbenp.prettier-vscode"
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,4 @@
 {
-  "debug.node.autoAttach": "on",
   "license-header-manager.excludeExtensions": [
     ".sh",
     ".js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
   },
   "scripts": {
     "build": "lerna run build",
-    "clean": "lerna run clean && rimraf yarn.lock && rimraf node_modules",
+    "clean": "lerna run clean && rimraf yarn.lock && rimraf node_modules; jest --clearCache",
```
jonnyhork, line 30: I'm not sure if we want this to also be in a script of its own or its enough to just have it in the package json somewhere so that we remember what the command is??

.vscode/settings.json
```diff
@@ -60,5 +60,7 @@
     "activityBar.background": "#523d89",
     "titleBar.activeBackground": "#523d89",
     "titleBar.activeForeground": "#f3f1ef"
-  }
+  },
+  "prettier.configPath": "../.prettierrc",
```
jonnyhork, line 64: I am also not sure if this is the correct path. The file is one level deep in `.vscode` but maybe the path is supposed to be from the top level of the repo instead of the file?

package.json
jgellin-sf: I think it's fine here. I don't think it makes much of a difference in this case, but all the other commands in the `clean` script are separated by `&&` and this one is separated by `;`. Maybe change to `&&` for consistency?

Here's a good explanation of the difference between those: https://unix.stackexchange.com/questions/100704/difference-between-executing-multiple-commands-with-and

jgellin-sf (commented):
Just the one comment about command separators in the `clean` script. It's nothing I'd hold up merging for.

package.json
jonnyhork: Sounds good to me, fixed in the `use consistant logical operator` commit.

jgellin-sf (approved):

PR #164, opened 2021-02-03
Track to force render on condition change

jgellin-sf:
### What does this PR do?
When there is only one condition in the Filter section of the SOQL builder, and it is removed with the X button, the condition inputs should be cleared. This should be triggered when the app externally sets the condition to the default condition template value (no field, '=' operator, no value). The condition is set appropriately, but none of the display values that were being tracked in order to trigger re-rendering of the component. This PR adds a `@track` annotation to the `_criteriaDisplayValue` to force the component to render when the condition is set.

@W-8827875@


dehru (approved):

jonnyhork (changes requested):
The value of the field is not cleared when the model changes,  such as when the entire condition is removed. All other values are cleared except the field.

PR #168, opened 2021-02-04
Jhork/clear modifier field bug fix

jonnyhork:
### What does this PR do?
This PR will address a bug where the field of the last modifier group would not be cleared because the customSelect component would not update the _value with an empty array coming from the model.

I have also refactored a skipped test to work with the customSelect.
@W-8847700@

### DIFF
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
@@ -48,7 +48,7 @@ export default class CustomSelect extends LightningElement {
 
   set selectedOptions(selectedOptions: string[]) {
     this._selectedOptions = selectedOptions || [];
-    if (selectedOptions && selectedOptions.length) {
+    if (selectedOptions) {
       this._value = selectedOptions;
     }
   }
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
@@ -47,10 +47,7 @@ export default class CustomSelect extends LightningElement {
   }
 
   set selectedOptions(selectedOptions: string[]) {
-    this._selectedOptions = selectedOptions || [];
-    if (selectedOptions) {
-      this._value = selectedOptions;
-    }
+    this._selectedOptions = this._value = selectedOptions || [];
   }
 
   @api
--- a/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.test.ts
+++ b/packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.test.ts
@@ -146,6 +146,35 @@ describe('WhereModifierGroup should', () => {
     expect(handler).toHaveBeenCalled();
   });
 
+  it('clears all the values when the X is clicked', () => {
+    modifierGroup.condition = {
+      field: { fieldName: 'foo' },
+      operator: '!=',
+      compareValue: { type: 'STRING', value: "'HELLO'" }
+    };
+    document.body.appendChild(modifierGroup);
+
+    const clearConditionBtn = modifierGroup.shadowRoot.querySelector(
+      '[data-el-where-delete]'
+    );
+    const {
+      selectFieldEl,
+      selectOperatorEl,
+      criteriaInputEl
+    } = getModifierElements();
+
+    expect(selectFieldEl.value[0]).toEqual('foo');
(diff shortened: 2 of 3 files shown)


jgellin-sf (approved):
👍

jgellin-sf:
I attempted to QA and there is a problem. The field value is tied to the condition model. When there is a validation problem the component is re-rendered to show the error, but the condition is not set. So on re-render the field is read from the condition (knock knock; nobody home) and the field input is reset. Here's a video:

https://user-images.githubusercontent.com/55159130/106913855-72325a80-66d2-11eb-9f22-c9fa0e5df212.mov

I think we need to decouple the custom select display value for the field from the condition model. I'll give that a whorl this morning.

jgellin-sf:
I fixed the issue I described with the `De-couple display field from condition model` commit. However there is still another problem. If there is an incomplete condition (all modifiers do not have value OR validation fails so the condition is set), the inputs are not cleared.


https://user-images.githubusercontent.com/55159130/106916125-d5bd8780-66d4-11eb-8d24-532a5051d694.mov

I think we need to explicitly clear the display values on X click. Back to work.

jgellin-sf:
Ok one more commit `Explicitly clear inputs on X click` to fix that last issue. I think we need an independent QA on this one.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -122,18 +129,8 @@ export default class WhereModifierGroup extends LightningElement {
...
   get _selectedField() {
-    return this.getFieldName() ? [this.getFieldName()] : [];
-  }
-
-  get filteredFields() {
-    return this.allFields.filter((field) => {
-      return field !== this.getFieldName();
-    });
+    return this._currentFieldSelection ? [this._currentFieldSelection] : [];
   }
```
jonnyhork, line 134: This is dead code, I forgot to remove this when we switched to the custom select

jonnyhork (commented):
How do we feel about testing to cover these sort of cases? I think we have the individual parts, clear modifiers, render errors, etc, each with a happy path. Do we want something that combines these conditions? Reset with and without errors.

I'm not sure if the issues we found here should have been caught by tests?

jonnyhork:
One more thing: If there was no condition in the model, such as when the user is filling out the first modifier group and makes a field selection but then tries to clear it, `_currentFieldSelection` is redefined but the UI is not updated. I added `@track` to `_currentFieldSelection` to cover this.

https://user-images.githubusercontent.com/25188632/106923457-52e3ff00-66cb-11eb-9430-63d4722a652a.mov

packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.ts
```diff
@@ -48,7 +48,7 @@ export default class CustomSelect extends LightningElement {
 
   set selectedOptions(selectedOptions: string[]) {
     this._selectedOptions = selectedOptions || [];
-    if (selectedOptions && selectedOptions.length) {
+    if (selectedOptions) {
```
dehru, line 51: Now that the length is removed, this can be simplified to.

```
set selectedOptions(selectedOptions: string[]) {
    this._selectedOptions = this._value = selectedOptions || [];
  }
```

jonnyhork: cool, fixed in the `simplify syntax` commit.

jonnyhork:
I don't see any tests that actually click the clear modifier group `X` and see that it works to clear the values, only send and event. I'll work on that.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.test.ts
```diff
@@ -146,7 +146,36 @@ describe('WhereModifierGroup should', () => {
     expect(handler).toHaveBeenCalled();
   });
 
-  it.skip('updates inputs when condition model changes', () => {
+  it('clears all the values when the X is clicked', () => {
```
jonnyhork, line 149: Do you think it is worth having a similar test but when there is an error state, like in 
`set error class on invalid operator input`

dehru: Yes, i'm adding that now.  It looks like the error state needs to be cleared as well in this case.

dehru: That'll be in a separate PR.  QAing this now.

dehru (approved):

PR #169, opened 2021-02-04
do not send event when user still adding input'

dehru:
### What does this PR do?
This PR fixes a bug in the multi-value input handler.  Such that the user is in the middle of typing a comma separated list of values, if the string ends in a comma and optional space(s), then the handler does not trigger a change event.

BEFORE

https://user-images.githubusercontent.com/599418/106939053-fdb0e900-66dc-11eb-8815-72fd5d10a890.mov

AFTER

https://user-images.githubusercontent.com/599418/106939017-effb6380-66dc-11eb-8a0c-282832ddcd08.mov

@W-8850622@


jgellin-sf (approved):
LGTM

PR #174, opened 2021-02-04
using svg for chevron to work on windows

dehru:
### What does this PR do?
This uses an svg for the custom drop-down chevron because on windows the font that is being used to render the character ^ is super skinny, and does not look anything like the standard select dropdowns with the down arrow.

BEFORE ON WINDOWS.
Screenshot 2021-02-04 124410

AFTER ON WINDOWS.
https://user-images.githubusercontent.com/599418/106965041-4e860900-6700-11eb-87b4-8f4d8c9c05cb.mov

AFTER ON MAC
Screen Shot 2021-02-04 at 3 44 36 PM
 
@W-8852383@


packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.html
```diff
@@ -31,7 +31,21 @@
...
+            width="8"
+            height="8"
+            x="0"
+            y="0"
+            viewBox="0 0 330.002 330.002"
+            style="enable-background: new 0 0 330.002 330.002"
+            xml:space="preserve"
+          >
+            <path
+              fill="white"
```
jonnyhork, line 44: What happens when you change to a light color theme? I previously had trouble styling the SVG the way I wanted. 
Currently the arrow changes from white to black depending on the theme like the rest of the text I think.

jonnyhork: Actually it looks like we have a mixed bag of colors depending on the theme
Icon colors

dehru:
I'm fixing this in the PR to remove the magnifying glass.  @jonnyhork

PR #177, opened 2021-02-05
Select COUNT()

jgellin-sf:
### What does this PR do?
This PR adds COUNT() to the SOQL model and the SOQL builder fields selector. COUNT() and other fields are mutually exclusive, so when COUNT() is selected, all other selections are cleared. When some other field is selected, COUNT() is removed.

https://user-images.githubusercontent.com/55159130/107048442-f2ba8f00-6796-11eb-87cf-61c9dc7e3a63.mov

@W-8786165@

### DIFF
--- a/packages/soql-builder-ui/src/modules/querybuilder/app/app.html
+++ b/packages/soql-builder-ui/src/modules/querybuilder/app/app.html
@@ -29,7 +29,6 @@
           selected-fields={query.fields}
           fields={fields}
           onfields__selected={handleFieldSelected}
-          onfields__removed={handleFieldRemoved}
           has-error={hasRecoverableFieldsError}
           is-loading={isFieldsLoading}
           class={theme}
--- a/packages/soql-builder-ui/src/modules/querybuilder/fields/fields.ts
+++ b/packages/soql-builder-ui/src/modules/querybuilder/fields/fields.ts
@@ -26,14 +26,16 @@ export default class Fields extends LightningElement {
   handleFieldSelection(e) {
     e.preventDefault();
     if (e.detail && e.detail.value) {
+      let selection = [];
       // COUNT() and other fields are mutually exclusive
       if (e.detail.value.toLowerCase() === 'count()') {
-        this.removeNonCountSelections();
+        selection.push('COUNT()');
       } else {
-        this.removeCountSelection();
+        selection = this.selectedFields.filter(value => value.toLowerCase() !== 'count()');
+        selection.push(e.detail.value);
       }
       const fieldSelectedEvent = new CustomEvent('fields__selected', {
-        detail: { field: e.detail.value }
+        detail: { fields: selection }
       });
       this.dispatchEvent(fieldSelectedEvent);
     }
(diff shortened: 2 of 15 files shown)


jonnyhork (dismissed):
Way to crank this out Jon. Looks good to me.

Event though the value of `count()` is not likely to change, do you think it would be better to make it a constant on the UI side?

Also, I know this UX mirrors Workbench, and Im not exactly sure of what customers do mist often but I wish selecting `count()` was more of a toggle. One that saves the state of previous selections. I would be rather annoyed if I selected say 15 fields and then hit count and they all clear, so I have to re-select the fields one by one.

packages/soql-builder-ui/src/modules/querybuilder/app/app.html
```diff
@@ -29,7 +29,6 @@
           selected-fields={query.fields}
           fields={fields}
           onfields__selected={handleFieldSelected}
```
jgellin-sf, line 31: The fields component now has more responsibility for maintaining it's selection, and only sends a single type of message, containing its entire selection, rather than the selection being maintained elsewhere via add and remove messages. The reason for the consolidation is that  multiple changes (such as removing existing selected fields to add 'COUNT()') were resulting in multiple messages, and integration with VSCode would not work because VSCode debounces messages, so only one message would get through.

dehru:
GEEK TIME:
This really highlights the direction of data flow in the UI, which is by design, but interesting still to geek out on.    

Let's say the fields were rendered with `SELECT Id, Name from Account`.  Then the user selects `COUNT()`.  For a millisecond, the rendered state of the fields component and the state of the model are out of sync.   The fields component state represents the previous state which is this.selectedFields = ['Id', 'Account'].  The event with the new user selectedFields being 'COUNT()' is still in flight, but represents a future state.  That event is caught in app and processed in toolingModelService, which updates the immutableModel.  App detects that model change and then sets the selectedFields = ['COUNT()'].  Only then, does the rendered UI match what the user actually selected.  

Luckily that is all super fast.

Our ui data flow concepts:
1. one-way data flow
2. state is read only
3. the model is the single source of truth
4. UI components are stateless ( mostly, although whereModifierGoup holds some temporary state when editing )
5. Changes to the model are made with pure functions ( immutableJS )

This mirrors REDUX, or more specifically FLUX ( because we use multiple stores in RxJS ).

packages/soql-builder-ui/src/modules/querybuilder/fields/fields.ts
```diff
@@ -8,27 +8,46 @@
...
+      // COUNT() and other fields are mutually exclusive
+      if (e.detail.value.toLowerCase() === 'count()') {
+        selection.push('COUNT()');
+      } else {
+        selection = this.selectedFields.filter(value => value.toLowerCase() !== 'count()');
+        selection.push(e.detail.value);
+      }
       const fieldSelectedEvent = new CustomEvent('fields__selected', {
-        detail: { field: e.detail.value }
+        detail: { fields: selection }
```
dehru, line 38: When multiple field selection is implemented, we are going to have to decide what to do if user selects multiple fields, once being COUNT().

dehru (changes requested):
This is pretty cool.  I enjoyed digging through this.  I agree with Jonny, the string COUNT() appears 65 times between soql-model and soql-builder-ui.  Most of those are appropriate without using a constant, like being used in soql statements for tests.  But a few of them, especially in the UI could benefit from strategic constantization ( LOL i made that up ).

jgellin-sf:
@dehru `COUNT()` is now constantizated.

dehru:
@jonnyhork - RE: The feature about reselecting the previous fields you mentioned.  Check this out.  Undo to the rescue.  

https://user-images.githubusercontent.com/599418/107586614-80dead00-6bbd-11eb-8572-0e9f16440052.mov

dehru (approved):

dehru (approved):

PR #179, opened 2021-02-05
Dehru/reset error flags on x

dehru:
### What does this PR do?

This clears the tooltip errors when the user clicks the 'X'.  Which fixes a bug where the tooltip error remains when the last WHERE clause is removed.

@W-8855869@

BEFORE

https://user-images.githubusercontent.com/599418/107066833-b2a9db00-679b-11eb-9334-6670f45b4bce.mov

AFTER

https://user-images.githubusercontent.com/599418/107067051-f997d080-679b-11eb-9631-4d6064089486.mov


dehru:
Oops, i need to redo this off of develop.  I'll fix the commit history.

PR #192, opened 2021-02-16
Jhork: Better Support For LIKE Operators

jonnyhork:
### What does this PR do?
This PR will add support for special LIKE operators:
**_default_**
`like` ---> abc
**_Special cases_**
`starts with` ---> abc%
`ends with` ---> %abc
`contains` ---> %abc%

If there is any existing wildcards (%) between any non-wildcard characters they will be ignored, otherwise any wildcards on the left or right of the value with be trimmed and replaced with a single wildcard in the appropriate place.

**Non functional changes:**
- refactored some existing code in soql-model `model.ts` and `validator.ts` to use and share constants
- organized some of the code in `whereModifierGroup.ts`

https://user-images.githubusercontent.com/25188632/108135120-c515f580-7074-11eb-94e5-3919b03e15a8.mov

@W-8555421@


packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
```diff
@@ -278,3 +278,79 @@ export function displayValueToSoqlStringLiteral(displayString: string): string {
...
+/* LIKE_CONTAINS %ABC% */
+export function isLikeContains(value: string) {
+  if (value && value.length) {
+    value = soqlStringLiteralToDisplayValue(value);
+    if (value.startsWith(WILD_CARD) && value.endsWith(WILD_CARD)) {
+      return true;
+    }
+  }
+  return false;
+}
```
jonnyhork, line 314: At first I was also including some check on the number of wild cards in the value, if its one of these special cases then there should be either one or two wildcards. However, since we want allow the user to have wildcard characters in between non-wildcard characters, `LIKE '100% Wild'`, then counting became tricky so now I will only look at the two outside characters to make a determination.

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.ts
```diff
@@ -278,3 +278,79 @@ export function displayValueToSoqlStringLiteral(displayString: string): string {
...
+      value = `${WILD_CARD}${value}`;
+      break;
+    case Soql.UiOperatorValue.LIKE_CONTAINS:
+      value = `${WILD_CARD}${value}${WILD_CARD}`;
+      break;
+    default:
+      break;
+  }
+  return value;
+}
```
jonnyhork, line 335: We will strip any existing wildcards form the outside of the condition value before adding back in one or two `%` in the appropriate place.
`%%Hey%%%` ---> starts with ---> `Hey%`

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -90,21 +93,40 @@ export default class WhereModifierGroup extends LightningElement {
...
-        condition.compareValue.value
+        condition.compareValue.value &&
+        matchingOption &&
+        matchingOption.value
       ) {
         this._criteriaDisplayValue = this.displayValue(
           condition.compareValue.type,
-          condition.compareValue.value
+          condition.compareValue.value,
+          matchingOption.value
```
jonnyhork, line 103: We need to pass in the type of operator that is selected so that we can handle some situations differently, currently just LIKE conditions.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -250,7 +256,12 @@ export default class WhereModifierGroup extends LightningElement {
         default: {
           // treat like string
           if (value.toLowerCase().trim() !== 'null') {
-            normalized = displayValueToSoqlStringLiteral(value);
+            if (this.isSpecialLikeCondition(operatorValue)) {
+              let wildCardValue = addWildCardToValue(operatorValue, value);
```
jonnyhork, line 260: This is where we add in the wildcard `%` to the compare value in the query string.

packages/soql-builder-ui/src/modules/querybuilder/whereModifierGroup/whereModifierGroup.ts
```diff
@@ -319,7 +330,11 @@ export default class WhereModifierGroup extends LightningElement {
 
       this._criteriaDisplayValue = this.criteriaEl.value;
       const type = this.getSObjectFieldType(fieldName);
-      const normalizedInput = this.normalizeInput(type, this.criteriaEl.value);
+      const normalizedInput = this.normalizeInput(
+        type,
+        this.criteriaEl.value,
+        op
```
jonnyhork, line 336: `op` is going to `UiOperatorValue` such as `EQ`, `LIKE`, `LIKE_START` etc

packages/soql-model/src/model/model.ts
```diff
@@ -64,6 +63,24 @@ export enum SObjectFieldType {
   Url = 'url'
 }
 
+export enum UiOperatorValue {
```
jonnyhork, line 66: While these values are primarily for the UI, they are consumed in both the soql-model and builder-ui. I thought it would be better to export this _from_ the `soql-model` since it is already a dependency of `soql-builder-ui`.

packages/soql-model/src/validators/validator.ts
```diff
@@ -30,58 +29,59 @@ export class DefaultValidator extends Validator {
...
-  { description: OPERATOR_NOT_IN, display: 'NOT IN', types: [SObjectFieldType.Address, SObjectFieldType.AnyType, SObjectFieldType.Base64, SObjectFieldType.Boolean, SObjectFieldType.Combobox, SObjectFieldType.ComplexValue, SObjectFieldType.Currency, SObjectFieldType.Date, SObjectFieldType.DateTime, SObjectFieldType.Double, SObjectFieldType.Email, SObjectFieldType.EncryptedString, SObjectFieldType.Id, SObjectFieldType.Integer, SObjectFieldType.Location, SObjectFieldType.Long, SObjectFieldType.MultiPicklist, SObjectFieldType.Percent, SObjectFieldType.Phone, SObjectFieldType.Picklist, SObjectFieldType.Reference, SObjectFieldType.String, SObjectFieldType.TextArea, SObjectFieldType.Time, SObjectFieldType.Url] },
-  { description: OPERATOR_INCLUDES, display: 'INCLUDES', types: [SObjectFieldType.MultiPicklist] },
-  { description: OPERATOR_EXCLUDES, display: 'EXCLUDES', types: [SObjectFieldType.MultiPicklist] }
+// prettier-ignore
```
jonnyhork, line 37: Keeps the array printed in one line instead of vertically. This makes it easier to scroll through this file.

packages/soql-model/src/validators/validator.ts
```diff
@@ -30,58 +29,59 @@ export class DefaultValidator extends Validator {
...
+  { description: UiOperatorValue.GT, display: '>', types: [SObjectFieldType.Address, SObjectFieldType.AnyType, SObjectFieldType.Base64, SObjectFieldType.Combobox, SObjectFieldType.ComplexValue, SObjectFieldType.Currency, SObjectFieldType.Date, SObjectFieldType.DateTime, SObjectFieldType.Double, SObjectFieldType.Email, SObjectFieldType.EncryptedString, SObjectFieldType.Id, SObjectFieldType.Integer, SObjectFieldType.Location, SObjectFieldType.Long, SObjectFieldType.MultiPicklist, SObjectFieldType.Percent, SObjectFieldType.Phone, SObjectFieldType.Picklist, SObjectFieldType.Reference, SObjectFieldType.String, SObjectFieldType.TextArea, SObjectFieldType.Time, SObjectFieldType.Url] },
+  { description: UiOperatorValue.LIKE, display: 'LIKE', types: LIKE_TYPES },
+  { description: UiOperatorValue.LIKE_START, display: 'LIKE', types: LIKE_TYPES },
+  { description: UiOperatorValue.LIKE_END, display: 'LIKE', types: LIKE_TYPES },
+  { description: UiOperatorValue.LIKE_CONTAINS, display: 'LIKE', types: LIKE_TYPES },
```
jonnyhork, line 51: Even though there are separate operators in the UI, they are all just `LIKE` as far as the model is concerned. Adding the special cases (starts, ends, etc) independently to this list was the easiest way to get the same validation as `LIKE`

jonnyhork (commented):
A lot of the diff here is formatting and or me re-arranging some of the code. I tried to highlight most of my changes with comments.

jgellin-sf (commented):
I think the code looks good. The introduction of the new operators makes me wonder about the order in which they appear in the drop-down. The order is now:
=
≠
<
≤
>
≥
in
not in
includes
excludes
like
starts with
ends with
contains

@smaddox-sf What do you think of the order? My inclination would be to move the various LIKEs above the multi-value operators (IN, INCLUDES, etc.).

jonnyhork:
> =
> ≠
> <
> ≤
> 
> > 
> 
> ≥
> in
> not in
> includes
> excludes
> like
> starts with
> ends with
> contains
> 
> @smaddox-sf What do you think of the order? My inclination would be to move the various LIKEs above the multi-value operators (IN, INCLUDES, etc.).

@jgellin-sf I'm curious what you are considering when thinking about the order? Is it what people will use most often, based on relation to other operators, purely esthetic or something else?

jgellin-sf:
> @jgellin-sf I'm curious what you are considering when thinking about the order? Is it what people will use most often, based on relation to other operators, purely esthetic or something else?

@jonnyhork I think I'm going to answer this in the least satisfying way possible... it _feels_ right to me to group the LIKE variations up with the other single-value operators. While I don't have numbers, I would assume = is by far the most used operator and INCLUDES/EXCLUDES (limited to multi-select picklists) are the least used. IN and NOT IN are similar in how they're expressed to INCLUDES/EXCLUDES, so I group those four together at the end. So I would say it's a combination of aesthetics and frequency, but also some other nebulous factor that I'll call _feel_.

jonnyhork:
haha, thats a legit answer! Its all about the feels!

packages/soql-builder-ui/src/modules/querybuilder/services/soqlUtils.test.ts
```diff
@@ -225,4 +231,71 @@ describe('SoqlUtils', () => {
       expect(actual).toEqual(expected);
     });
   });
+
+  describe('LIKE OPERATOR UTILS', () => {
```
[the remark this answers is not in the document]
dehru: that's awesome

dehru (approved):

PR #198, opened 2021-02-19
refactor: factor out common code into new soql-common package

dobladez:
Also include the obfuscated output of the `soql-parser` package as a subdirectory.

### What does this PR do?
Move common code (handling of SOQL comments and ANTLR soql-parser) into a separate package.

Also: includes the generated soql-parser as a regular directory and drop the tgz dependency.

W-8842328


jgellin-sf (commented):
Looks like a good start

dehru (approved):

PR #205, opened 2021-02-23
fix: update imports from soql-common

jonnyhork:
### What does this PR do?
This PR will update some imports from the soql-common package so the build does not fail.


dobladez (approved):
👍

jgellin-sf (approved):

PR #206, opened 2021-02-23
Jhork/update lwc services

jonnyhork:
### What does this PR do?
This PR will update the version of `lwc-services` to `v^3.0.2`, with this come Webpack 5 which will allow us to use the latest features in TS such as optional chaining.

However Webpack 5 no longer comes with [node module polyfills](https://github.com/webpack/webpack/pull/8460/commits/a68426e9255edcce7822480b78416837617ab065) by default, so we need to update our webpack config to satisfy `Antlr4ts`.

This PR will also update the base image used in CI to [`cimg/node:14.15.5`](https://circleci.com/developer/images/image/cimg/node)


packages/soql-builder-ui/package.json
```diff
@@ -18,10 +18,15 @@
...
     "eslint": "^7.5.0",
     "jest": "26.1.0",
-    "lwc-services": "2.2.1",
-    "prettier": "^2.0.5"
+    "lwc-services": "^3.0.2",
+    "os-browserify": "^0.3.0",
+    "prettier": "^2.0.5",
+    "process": "^0.11.10",
+    "terser-webpack-plugin": "^5.1.1",
+    "util": "^0.12.3"
```
jonnyhork, line 29: All of these are used by webpack as plugins for the node modules.

jonnyhork (commented):
There will  be a corresponding [PR](https://github.com/forcedotcom/salesforcedx-vscode/pull/2975) on the vs code side.

packages/soql-builder-ui/package.json
```diff
@@ -7,21 +7,25 @@
     "access": "public",
     "registry": "http://registry.npmjs.org/"
   },
-  "dependencies": {
-    "@salesforce/soql-model": "0.1.22",
-    "debounce": "^1.2.0",
-    "immutable": "3.8.2",
-    "rxjs": "^6.6.2"
-  },
+  "dependencies": {},
```
jonnyhork, line 10: I moved all the modules here to `devDependencies`. 
Because we are webpacking the application all the dependencies needed to run it are included in the `app.js` output form webpack. Previously, a consumer of `soql-builder-ui` would ultimately get runtime dependencies twice post-install because they we listed in the package.json of the package as well as in the output bundle of webpack.

dehru: This does lose information about what is actually used at runtime.  I suggest we add this.

```
"//dependencies": {
    "comments": "since these are all webpacked, they are moved to devDependencies",
    "@salesforce/soql-model": "0.1.22",
    ....
  }
```

dehru (approved):
I had a suggestion to keep some information about what is actually in the runtime.

packages/soql-builder-ui/package.json
jonnyhork: I tried adding comments, I don't think they are supported in the package.json, the build fails looking for `"comments":..` on npm.
I'm going to skip the comment for now.

dobladez: @jonnyhork may be you missed the`//` prefix Dehru put on "`//dependencies`" ?
For example, this worked for me:
```json
  "//dependencies": {
    "comments": "blah",
    "@salesforce/soql-model": "xyz"
  },
```

dobladez:
Note: Given that it now requires Node 14, we'd need to update the release workflow too (`.github/workflows/release-soql-builder-ui.yml`):
```
        with:
          node-version: 12
```

packages/soql-builder-ui/package.json
jonnyhork: yeah guess so, I even google "comments in package.json" and didn't see that syntax...

ok I'll add it back in. Thanks

jonnyhork: Curious though, is this really useful to a consumer of the package?

jonnyhork: Also especially with version of the packages likely out of date by the time they look at it?

jonnyhork:
I updated the node version in the github workflow and added the comments back, thanks for the feedback.

packages/soql-builder-ui/package.json
dehru: @jonnyhork - it’s really just for us.  I always find it important for devs to know the difference between what modules are loaded into the browser vs what’s just running in the node tooling.

when we move everything to devDependencies we lose that information.  Especially think about a new dev joining the team.

jonnyhork: Isn't this case isn't all of it running in the browser per-se (webview), since its all webpacked?

dehru: Nah, webpack starts here at soql-builder-ui/src/index.ts ( js actually ) and begins packing only references it finds.  The dependency tree related to imports starting at that file.  

https://github.com/forcedotcom/soql-tooling/blob/main/packages/soql-builder-ui/src/index.ts

It’s a huge graph ( or tree of import/require statements  ).  But things like jest, prettier, eslint, and the typescript types do not get webpacked because they are never imported into the LWC web app.

dobladez (approved):
Looks good. Once we release, we'll see how Node v14 works for our publishing script 🤞

PR #208, opened 2021-02-23
build: update internal dependency versions on release PRs

dobladez:
### What does this PR do?
Improve release scripts to automatically keep the version inter-dependencies updated on the release PRs.

Example: On the release PR for soql-model,  the script now updates soql-builder-ui's package.json to depend on the new version.

W-8700164

### DIFF
--- a/.github/workflows/release-soql-builder-ui.yml
+++ b/.github/workflows/release-soql-builder-ui.yml
@@ -20,21 +20,38 @@ jobs:
           path: packages/soql-builder-ui
           monorepo-tags: true
           default-branch: main
+      - run: |
+          echo 'Release outputs:${{ toJSON(steps.release.outputs) }}'
+        # Case 1:
+        # If this is not the result of a Release PR merge, then update version
+        # of internal dependencies on the release branch
       - uses: actions/checkout@v2
-        # these if statements ensure that a publication only occurs when
-        # a new release is created (when merging a release PR to main):
+        if: ${{ ! steps.release.outputs.release_created }}
+      - name: bump-inter-dependencies-on-release-branch
+        if: ${{ ! steps.release.outputs.release_created }}
+        run: ./scripts/gh-action-bump-interdeps-versions-on-release-branch.sh soql-builder-ui
+        env:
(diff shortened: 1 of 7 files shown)


praksb:
Do we really want to automate this or continue to keep this step manual? My concern is that if this would potentially prevent soql-model only changes to be blocked until the corresponding changes in soql-builder-ui is complete since otherwise there is a possibility of build break or functionality break in soql-builder-ui.

dobladez:
> Do we really want to automate this or continue to keep this step manual? My concern is that if this would potentially prevent soql-model only changes to be blocked until the corresponding changes in soql-builder-ui is complete since otherwise there is a possibility of build break or functionality break in soql-builder-ui.

That's a good point. What's the most common case? Is it to bump the deps together on release? or release independently?
May be we should default to the common case, leaving the option to manually override it.

Either way, having the separate "release" PRs allows us to override the changes before merging.

If we decide not to update the deps on the release PRs on every commit to `main`, one could run the same script locally, or we could also have the logic in a separate gh action and trigger it manually.

dehru:
We sure do a lot of work to keep a mono-repo.  Are we sure we don't wanna just have separate repos for:

- soql-builder-ui ( includes soql-model library )
- soql-data-view
- soql-common

jonnyhork:
I was just gonna say the same thing as @dehru 
How badly or how soon do we want to convert soql-model to a library in builder-ui? Because once we do that, then this automation wont be needed right?

praksb:
It seems like we are complicating our automation. Since soql-model is used only by the soql-builder-ui why do we even have to publish soql-model to npm? Instead, can we just take a source dependency on this like @jonnyhork suggested.

dobladez:
Sounds good. If we all agree on getting rid of the monorepo, let's do it :-). We can split soql-tooling into 3 separate repos like @dehru mentioned.

I just realized VSCode extension depends on soql-model. We might be able to refactor to cut that dependency.

dobladez:
Note: the workflow files (under `.github/workflows/*`) are all pretty much the same: only the package name changes. There's not built-in mechanism to parameterize and reuse gh workflows.

jonnyhork (approved):

PR #213, opened 2021-02-24
fix: super simple basic input maxlength controls

dehru:
### What does this PR do?
In honor of today's threat model conference, i implemented the most basic of html form input controls...maxlength.  And these limits i added are just a sanity upper limit, to prevent some kinda of performance degradation with vscode or possible "network congestions / buffer overflow" issue with the api. 

The threat model here is awkward, since pasting like 10,000 characters into the where criteria would likely only cause issues with vscode.  And the salesforce soql api that we query is well protected.  At least this would prevent a HUGE api call with some giant WHERE criteria. 

These controls aren't testable, since we don't actually type into the input elements during the tests, these limits can be broken with javascript.

For the @salesforce/soql-data-view module, the Tabulator library handles field display sanitation already. 

>  "To guard against code injection, any formatters that are meant to display text (plaintext, textarea, money, email, link) have the data sanitized first to escape any potentially harmful HTML or JavaScript from making into the table, this causes any such data to be displayed as its plain text alternative. "

See [Built In Formatters](http://tabulator.info/docs/4.9/format).

### DIFF
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.html
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.html
@@ -11,7 +11,7 @@
         onkeydown={handleKeyDown}
         onfocus={handleInputFocus}
         onblur={handleInputFocus}
-        maxlength="100"
+        maxlength="40"
       />
       <template if:true={hasSearchTerm}>
         <div
--- a/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.html
+++ b/packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.html
@@ -11,6 +11,7 @@
         onkeydown={handleKeyDown}
         onfocus={handleInputFocus}
         onblur={handleInputFocus}
+        maxlength="100"
       />
       <template if:true={hasSearchTerm}>
         <div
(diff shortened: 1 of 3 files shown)


packages/soql-builder-ui/src/modules/querybuilder/customSelect/customSelect.html
```diff
@@ -11,6 +11,7 @@
         onkeydown={handleKeyDown}
         onfocus={handleInputFocus}
         onblur={handleInputFocus}
+        maxlength="100"
```
jonnyhork, line 14: Do we really need one here since the input is only matching array values in memory? Also I cant verify but I bet there are some really long field names out there in customer orgs. Maybe not 100 characters though?

jgellin-sf: Maybe 40? https://help.salesforce.com/articleView?id=000313503&type=1&mode=1

dehru: yeah, it doesn't really help.  but i thought i'd stick it in anyway since it might get a false positive on a security scan.

jonnyhork: oh, yeah, good call on the security scan

jonnyhork (approved):

PR #214, opened 2021-02-25
fix: do not publish unnecessary files

jgellin-sf:
### What does this PR do?
This PR reduces the size of the SOQL Extension by not publishing unnecessary files in the dependencies.

@W-8592633@

### DIFF
--- a/packages/soql-builder-ui/.npmignore
+++ b/packages/soql-builder-ui/.npmignore
@@ -11,6 +11,7 @@ node_modules
 jest.config.js
 lwc-services.config.js
 lwc.config.json
+webpack.config.js
 
 # MacOS system files
 .DS_Store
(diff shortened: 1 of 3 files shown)


dobladez (approved):
👍

dehru (approved):
woot!  

It's unfortunate we can't do something like extending an existing config.  Kindalike we do with tsconfig.

jonnyhork:
Do we know how much this reduces the package size?

jonnyhork (approved):

jgellin-sf:
The move to putting the parser directly in `soql-common` from  bundling it as an external dependency was huge (thanks @dobladez) as it prevents us from including multiple copies of ANTLR, and should lead to a reduction of over 1 MB. The changes in this PR are easy to make and are on the order of 3KB to 50KB improvement per package, but every little bit helps I suppose. Once published I can build the VSIX and get an accurate metric on the entire VSIX.

There are a few additional things we can do, but those are more radical interventions that change the way we package (or webpack) things. Those interventions will need more discussion.

PR #218, opened 2021-02-26
build: set npm registry explicitly to fix npm publish auth errors

dobladez:
### What does this PR do?
Configures Github action "setup-node" with an explicit npm registry url. Otherwise, it seems to not take our NPM_TOKEN.

W-8700164


jgellin-sf (commented):
Don't we also need this here: https://github.com/forcedotcom/soql-tooling/blob/build/set-npm-registry-explicity/.github/workflows/release-soql-builder-ui.yml ?

jgellin-sf (approved):

dehru:
How the heck did you figure this out?

PR #231, opened 2021-03-04
fix: Support Build/Run in Windows Environment

jonnyhork:
### What does this PR do?
This PR is **PART 1/2** to better support soql-tooling in Windows. These changes focus on _building and running_ the SOQL Tooling repo.

There were some npm scripts that used commands or paths that are not cross-platform compatible. So we can use [`shx`](https://www.npmjs.com/package/shx) just like in the salesforcedx-vscode allow scripts to run on any platform.

With these changes I was able to build and run the builder UI on localhost:3001 in Windows:
SOQL Tooling in Windows

**Part 2** will focus on running our CI in a windows environment.

NOTE: When testing the `yarn clean` script I noticed and error coming form jest:
```bash
Validation Error:
  Preset @lwc/jest-preset not found.
  Configuration Documentation:
  https://jestjs.io/docs/configuration.html
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
Some investigation suggested that bumping the version of jest would fix this or changing the jest-config preset from `@lwc/jest-preset` to `./node-modules/@lwc/jest-preset` would help. The error is also present in `main`. I tried both but it did not fix it. From what I can tell the error is not causing any issues but I wanted to call it out incase it comes up in QA.
@W-8914907@


package.json
```diff
@@ -17,13 +17,13 @@
...
     "husky": "4.3.8",
     "lerna": "^3.19.0",
-    "rimraf": "^3.0.2",
+    "shx": "^0.3.3",
     "tslint": "^5.8.0",
     "typescript": "^4.0.3"
   },
   "scripts": {
     "build": "lerna run build",
-    "clean": "lerna run clean && rimraf yarn.lock && rimraf node_modules && jest --clearCache",
```
jonnyhork, line 26: `rimaf` would produce a `rimaf is not a recognized command` error in Windows 10. As far as I can tell its just a sub for `rm -rf` so thats what I used.

jonnyhork (commented):
shx is the way to go for simple scripts in the package.json, however anything more complex and we should consider using `sheljs` instead of bash scripts. I think it will be easier to work with across different platforms.

dehru (approved):

dehru:
Since nobody else can test this on windows....I guess we can't QA it on windows.  ;-)

dehru:
I tested this  branch on mac, everything still works.

jgellin-sf (approved):

PR #235, opened 2021-03-10
chore: Run soql-tooling CI in Windows

jonnyhork:
### What does this PR do?
This PR will 
1. reconfigure our Circle CI config to run in a Windows Environment. 
1. update some of our test to use cross-platform End-Of-Line characters with the `os` library.
@W-8914907@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,71 +1,91 @@
 version: 2.1
 
 orbs:
+  win: circleci/windows@2.2.0
   slack: circleci/slack@3.4.2
 
-variables:
-  - &node-build-image
-    - image: cimg/node:14.15.5
+perameters:
+  node-version:
+    type: string
+    default: '14.15.5'
+
+commands:
+  build-windows:
+    steps:
+      - run:
+          name: 'Install node'
+          shell: bash.exe
+          command: |
+            nvm install $NODE_VERSION
+            nvm use $NODE_VERSION
+            node --version
+            yarn --version
+      - run:
+          name: 'Install dependencies'
+          shell: bash.exe
+          command: yarn
+      - run:
+          name: 'Compile Project'
+          shell: bash.exe
+          command: |
+            echo 'Build all packages for linked dependencies'
+            yarn build
 
 jobs:
   soql-builder-ui:
-    docker: *node-build-image
+    executor: win/default
     environment:
       PACKAGE_PATH: 'packages/soql-builder-ui'
+      NODE_VERSION: << pipeline.parameters.node-version >>
     steps:
       - checkout
       - restore_cache:
           key: dependency-cache-{{ checksum "yarn.lock" }}
-      - run: yarn
+      - build-windows
       - save_cache:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -1,71 +1,122 @@
...
+          shell: bash.exe
+          command: |
+            nvm install $NODE_VERSION
+            nvm use $NODE_VERSION
+      - run:
+          name: 'Install yarn'
+          shell: bash.exe
+          command: |
+            npm install --global yarn
+            yarn --version
```
jonnyhork, line 26: Unlike the language server [config](https://github.com/forcedotcom/soql-language-server/pull/21/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R32) I got an error when I tried to run `yarn` commands in the same `- run` block, so I needed to install yarn separately. This may be because I am using nvm to switch node versions.

.circleci/config.yml
```diff
@@ -1,71 +1,122 @@
...
       - run:
           name: Upload coverage report to Codecov
-          command: bash <(curl -s https://codecov.io/bash) -F soql-model -t ${CODECOV_TOKEN}
+          shell: bash.exe
+          command: bash <(curl -s https://codecov.io/bash) -F soql-builder-ui -t ${CODECOV_TOKEN}
 
   notify_slack:
-    docker: *node-build-image
+    docker:
+      - image: cimg/node:<< pipeline.parameters.node-version >>
```
jonnyhork, line 119: This slack job had issues running in Windows, it was looking for bash, even though I added the `shell: bash.exe` property. Defining a node env allowed it to run without errors.
There is a much newer version of the slack orb `4.3.0` that is supposed to have [better Windows support](https://github.com/CircleCI-Public/slack-orb/pull/160) however there are still a few pending issues in the repo related to Windows and I don't see need to "fix what isn't broken"

.circleci/config.yml
```diff
@@ -1,71 +1,122 @@
...
+            echo 'Run SOQL Model Unit Tests'
+            cd $PACKAGE_PATH
+            yarn run test:unit:coverage
       - run:
-          name: 'Run SOQL Model Server Unit Tests'
+          name: Upload coverage report to Codecov
+          shell: bash.exe
+          command: bash <(curl -s https://codecov.io/bash) -F soql-builder-ui -t ${CODECOV_TOKEN}
+
+  soql-common:
```
dobladez, line 91: Wow we didn't even have soql-common on CircleCI!  That was my fault 🤦

dobladez (approved):
Looks 👍

.circleci/config.yml
```diff
@@ -1,71 +1,122 @@
...
+    default: '14.15.5'
+
+commands:
+  build-windows:
+    steps:
+      - run:
+          name: 'Install node'
+          shell: bash.exe
+          command: |
+            nvm install $NODE_VERSION
```
dehru, line 19: So nvm is already installed?

dobladez: yes

PR #236, opened 2021-03-11
feat: provide accessible labels

dehru:
### What does this PR do?
This adds labels to some of the more complex soql builder components.  For instance, the triplet WHERE inputs and ORDER BY inputs.  According to the accessibility folks here at salesforce, we need to use hidden labels instead of the aria-label attribute.

@W-8984214@

Interestingly, LWC automatically adds unique id's to elements if you give them an id.  So i don't need to worry about id collision.

Screen Shot 2021-03-11 at 2 31 47 PM

### DIFF
--- a/packages/soql-builder-ui/package.json
+++ b/packages/soql-builder-ui/package.json
@@ -9,7 +9,7 @@
   },
   "//dependencies": {
     "comments": "since these are all webpacked, they are moved to devDependencies",
-    "@salesforce/soql-model": "0.1.22",
+    "@salesforce/soql-model": "current version in ../soql-model/package.json",
     "debounce": "^1.2.0",
     "immutable": "3.8.2",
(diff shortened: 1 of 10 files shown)


dehru:
I'm updating this as the accessibility team says that aria-label is not yet supported and we should use actual hidden labels.

jgellin-sf:
QA failed. The labels are successfully inserted into the DOM, but they are not being read by VoiceOver.

dehru:
We are punting on accessibility for now.

jonnyhork:
I'm curious why we wouldn't at least add the `aria-label` attributes? We saw that they worked for us with Voice Over and they do not effect function in any way. Even though the A11y team suggested a different approach , adding the `aria-labels` _only adds value_ in the meantime, and that could impact at least one or more people in the future.
CC @smaddox-sf @dehru

PR #237, opened 2021-03-12
build: update and enforce eslint and prettier rules

dobladez:
* run "lint" command (eslint+prettier rules) on git-push
* ~~"lint" command on soql-builder-ui and soql-model is disabled for now
  (renamed) because they currently have too many violations.~~
* ~~Might want to fix these violations incrementaly in the future.
  To this end, vscode workspace now configured to format "modified lines only"~~

W-8915011


dobladez:
Note: these rules are in our `.eslintrc.json` until  [this PR on sfdx-dev-packages](https://github.com/forcedotcom/sfdx-dev-packages/pull/119) is merged:

```
  "rules": {
    "no-shadow": "off",
    "@typescript-eslint/no-shadow": "error"
  }
```

.eslintrc.json
```diff
@@ -0,0 +1,10 @@
+{
```
dehru, line 1: Was there a repo you used as a base for this config?

packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
```diff
@@ -29,54 +29,60 @@ import { lwcIndexableArray } from '../services/lwcUtils';
...
-  messageService: IMessageService;
-  theme = 'light';
-  sobjectMetadata: any;
-  notifications = [];
-
-  get shouldBlockQueryBuilder() {
-    return ( this.hasUnrecoverableError || this.hasUnsupportedMessage ) && this.dismissNotifications === false;
-  }
-  get showUnsupportedNotification() {
+  public fields: string[] = [];
```
dehru, line 37: I'm not proposing any changes here.  Just mentioning that the melding of LWC @api annotations and typescript's public access modifier is an odd crossover.  In reality, these probably should just be `protected`.  And only set the ones annotated with @api to public, just to express the true intent.  However, i realize that having no modifiers is also essentially a `public` modifier.  

In the case of LWC components, the typescript access modifiers really have very little purpose except allow us to extend the class in a test and modify it's annotations so we can access it's internals.

packages/soql-builder-ui/src/modules/querybuilder/from/from.test.ts
```diff
@@ -1,3 +1,9 @@
+/* eslint-disable no-underscore-dangle */
```
dehru, line 1: I'd suggest we just disable this for the whole package.

packages/soql-builder-ui/src/modules/querybuilder/limit/limit.ts
```diff
@@ -9,15 +9,17 @@
 import { LightningElement, api } from 'lwc';
 
 export default class Limit extends LightningElement {
-  @api hasError = false;
-  @api limit;
+  @api public hasError = false;
+  @api public limit;
 
-  handleLimitChange(e) {
+  /* eslint-disable @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-member-access */
```
dehru, line 15: I'm guessing typescript wants a definition of the Custom Event payload structure here.  I can see that as valuable on the receiving end as well.

packages/soql-builder-ui/src/modules/querybuilder/services/model.ts
```diff
@@ -61,6 +63,7 @@ export interface OperatorOption {
   value: string;
   displayValue: string;
   modelValue: Soql.ConditionOperator;
+  // eslint-disable-next-line no-unused-vars
```
dehru, line 66: Is this just a failure of eslint to understand the interface is being used below?

packages/soql-builder-ui/src/modules/querybuilder/services/toolingModelService.test.ts
```diff
@@ -91,21 +94,23 @@ describe('Tooling Model Service', () => {
...
 
-      expect(query!.fields.length).toEqual(0);
+      expect(query.fields.length).toEqual(0);
 
       // set
       modelService.setFields([mockField1, mockField2]);
-      expect(query!.fields.length).toBe(2);
-      expect(query!.fields).toContain(mockField1);
-      expect(query!.fields).toContain(mockField2);
+      expect(query.fields.length).toBe(2);
```
dehru, line 105: this is helpful

packages/soql-model/src/model/impl/whereImpl.test.ts
```diff
@@ -21,7 +23,7 @@ describe('WhereImpl should', () => {
     expect(actual).toEqual(expected);
   });
   it('return condition preceded by WHERE keyword for toSoqlSyntax()', () => {
-    const expected = "WHERE field = 'abc'"
+    const expected = "WHERE field = 'abc'";
```
dehru, line 26: I worked on a typescript project where we removed the requirement for the semi-colon.  It was awesome.  Never had an issue in the 2 years i worked on it.

packages/soql-model/src/model/model.ts
```diff
@@ -32,7 +34,7 @@ export enum ErrorType {
   NOCOMPAREVALUE = 'NOCOMPAREVALUE',
   NOCOMPAREOPERATOR = 'NOCOMPAREOPERATOR',
   INCOMPLETEMULTIVALUELIST = 'INCOMPLETEMULTIVALUELIST',
-  UNEXPECTEDEOF = 'UNEXPECTEDEOF'
+  UNEXPECTEDEOF = 'UNEXPECTEDEOF',
```
dehru, line 37: I find it interesting that the standard salesforce config wants trailing commas.

dehru (approved):
This is another honorable brute-force work item performed by Fernando.  It'd be interesting to run create-lwc-app some 9 months later and see if some of this is now in the template.

Sometimes I prefer some more inference of types, I like to honor JSON and it's implied ambiguity.  However, I guess I can live with the addition of eslint-ignore rules when necessary.  

I do wish we didn't have so many already.  Or there was a better way to loosen the noose to find a happy medium.

Either way, this work is valuable, and a little annoying, but that's not Fernando's fault.

.eslintrc.json
dobladez: No... there's almost nothing in it other than extending `eslint-config-salesforce-typescript`.
The "rules" section is to workaround [this problem](https://github.com/forcedotcom/sfdx-dev-packages/pull/119)

packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
dobladez: I'm not sure I fully understand our comment @dehru.  In short:  I didn't want to change the semantics of existing code... so, if the access modifier was missing, I put `public` (the equivalent).

If some cases should become `protected`, great... lets change them afterwards. This is the benefit of having the linter in this case: it forces you to think about the access modifier defaults, which are often ignored.

packages/soql-builder-ui/src/modules/querybuilder/services/model.ts
dobladez: Yep... I just found a workaround for this one too:

    "no-unused-vars": "off",
    "@typescript-eslint/no-unused-vars": ["error"]

will push a fix. Thanks!

packages/soql-builder-ui/src/modules/querybuilder/from/from.test.ts
dobladez: sounds good

packages/soql-builder-ui/src/modules/querybuilder/app/app.ts
dehru: My comment is from a component perspective.  The LWC components public api is exposed not by access modifiers, but by the annotation @api.  So it's just strange to have `public`, when in an LWC component, those actually behave as `protected`.  

You did exactly what you should have done.  It's just a philosophical mismatch between typescript and ui component frameworks that i wish was more aligned.  There is a similar problem with Angular.

dehru (approved):

PR #239, opened 2021-03-31
fix: better align header labels with inputs

jonnyhork:
### What does this PR do?
This PR will fix the misaligned header labels in the where clause.

The CSS will be updated so that the general spacing of each header is controlled with flexbox but the specific alignment is controlled by styles for each header value.
Before: 
image (2)

After: 
Screen Shot 2021-03-31 at 12 03 01 PM

### DIFF
--- a/packages/soql-builder-ui/src/modules/querybuilder/where/where.css
+++ b/packages/soql-builder-ui/src/modules/querybuilder/where/where.css
@@ -30,18 +30,20 @@ header.header {
   display: flex;
   justify-content: space-around;
   margin-top: 5px;
-  padding-right: 2.5em;
 }
 
 .header__labels p {
-  margin: 0 0.5em 0 0;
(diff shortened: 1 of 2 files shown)


dobladez (approved):

packages/soql-builder-ui/src/modules/querybuilder/where/where.css
```diff
@@ -30,18 +30,20 @@ header.header {
...
 }
 
 .header__labels p {
-  margin: 0 0.5em 0 0;
+  margin: 0;
 }
 /* better align labels with correct elements below */
 .header__labels p.header__label__field {
-  margin-left: 45px;
+  margin-left: 1rem;
```
dehru, line 40: it's great moving to rem.

dehru (commented):
looks good.

PR #248, opened 2021-05-04
fix: change wording when Builder cannot handle complex query

dobladez:
### What does this PR do?
Add extra sentence to wording when Builder cannot handle complex query

W-9154690

### DIFF
--- a/packages/soql-builder-ui/src/modules/querybuilder/app/app.html
+++ b/packages/soql-builder-ui/src/modules/querybuilder/app/app.html
@@ -66,14 +66,16 @@
               <template if:true={showUnsupportedNotification}>
                 <!-- TODO: i18n -->
                 <strong>Your query contains statements that SOQL Builder doesn't currently
-                support.</strong><p>You can use a a text editor to edit the query. If you
+                support.</strong>
+                <p>You can use the 'Run Query' button to run your query as is.</p>
+                <p>Consider using a text editor to edit the query. If you
                 click Edit Query Anyway, SOQL Builder removes the unsupported syntax.</p>
               </template>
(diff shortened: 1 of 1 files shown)


dehru (approved):

packages/soql-builder-ui/src/modules/querybuilder/app/app.html
```diff
@@ -66,7 +66,9 @@
               <template if:true={showUnsupportedNotification}>
                 <!-- TODO: i18n -->
                 <strong>Your query contains statements that SOQL Builder doesn't currently
-                support.</strong><p>You can use a a text editor to edit the query. If you
+                support.</strong>
+                <p>You can use the 'Run Query' button to run your query as is.</p>
+                <p>You can use a a text editor to edit the query. If you
```
jonnyhork, line 71:
```suggestion
                <p>You can use a text editor to edit the query. If you
```
There is a typo here with 2 `a`s

jonnyhork: Maybe we want to say
```
You can also use a text editor...
```
Adding, "also", I think this will join the two statements better about their options.

dobladez: Good catch! that double "a" has been there for awhile and nobody caught it (not even Emily or Sonal) :-).

Regarding adding "also",  Sonal proposed changing the phrase to

     Consider using a text editor to edit the query

which: _"... I think that might make the “Edit Query Anyway” button more valid..."_.  So I'll do that unless somebody objects.

dobladez:
FYI: Changed the sentence to: `Consider using a text editor to edit the query` as suggested by Sonal

PR #251, opened 2021-05-07
feat: add options to clear and select all fields

dobladez:
### What does this PR do?
Add two options on the Field selection dropdown, to select all fields, and to clear all selections.
Also: the pills with the selected fields now have a max height and vertical scrolling.

@W-8908803


packages/soql-builder-ui/src/modules/querybuilder/fields/fields.ts
```diff
@@ -10,9 +10,18 @@
 
 import { LightningElement, api } from 'lwc';
 import { SELECT_COUNT } from '../services/model';
+
+const SELECT_ALL_OPTION = '* (ALL)';
```
ryanbrainard, line 14: When this gets rendered in the UI, the `*` ends up looking like a bullet point and looks a little strange sandwiched in between the `- Clear Selection -` and `COUNT()`. This definitely is not a blocker and I'm not sure what would be better (maybe just `ALL FIELDS`?), but in case we want to improve the visuals here.

image

packages/soql-builder-ui/src/modules/querybuilder/app/app.html
```diff
@@ -65,17 +68,29 @@
             <div class="warning-notification__message">
               <template if:true={showUnsupportedNotification}>
                 <!-- TODO: i18n -->
-                <strong>Your query contains statements that SOQL Builder doesn't currently
-                support.</strong>
-                <p>You can use the 'Run Query' button to run your query as is.</p>
-                <p>Consider using a text editor to edit the query. If you
-                click Edit Query Anyway, SOQL Builder removes the unsupported syntax.</p>
+                <strong
```
ryanbrainard, line 71: Do you have an example of a query that would cause this?

ryanbrainard (approved):
LGTM!

On a side note, reviewing this PR was my first time setting up my local dev environment and I had some problems getting the extension to launch using the instructions in the readme. I was finally able to get it to work by using `yarn link @salesforce/soql-builder-ui` instead of `npm link @salesforce/soql-builder-ui`, so guessing the `yarn link` in the `soql-tooling` packages requires `yarn link` on the `salesforcedx-vscode/packages/salesforcedx-vscode-soql` side as well. Should we update both sides to use `yarn` in the readme?

packages/soql-builder-ui/src/modules/querybuilder/app/app.html
dehru: Right now we don't support display or building Joins in the UI

dehru:
> LGTM!
> 
> On a side note, reviewing this PR was my first time setting up my local dev environment and I had some problems getting the extension to launch using the instructions in the readme. I was finally able to get it to work by using `yarn link @salesforce/soql-builder-ui` instead of `npm link @salesforce/soql-builder-ui`, so guessing the `yarn link` in the `soql-tooling` packages requires `yarn link` on the `salesforcedx-vscode/packages/salesforcedx-vscode-soql` side as well. Should we update both sides to use `yarn` in the readme?

Ryan, I've always used npm link to integrate these, basically because the soql extension repo uses npm.  So I'm interested in where we missed some critical bit in the Readme.  Did you link from within packages/salesforcedx-vscode-soql?

dehru (dismissed):

jgellin-sf (dismissed):

packages/soql-builder-ui/src/modules/querybuilder/fields/fields.ts
ryanbrainard: Not sure what the procedure is there, but I opened a PR with the suggestion: https://github.com/forcedotcom/soql-tooling/pull/253

jonnyhork: Thank you for doing that! Typically I think we would just commit those changes to this PR so there is less to keep track of.

dehru (approved):

Issue #254, opened 2021-05-10
npm run lint fails with prettier errors


ryanbrainard:
**OS and version**:
MacOS 11.3

**VS Code Version**:
`main` (7a38048ed7c6d52f5eb663cf4b1cb5b6dc4bf12c)

**SFDX CLI Version**:
n/a

### Summary

`npm run lint` fails with prettier errors

### Steps To Reproduce:

1.  `npm run lint`

### Expected result

exit 0

### Actual result

```
─$ npm run lint

> @ lint /Users/rbrainard/Development/soql-tooling
> lerna run lint

lerna notice cli v3.22.1
lerna info Executing command in 3 packages: "yarn run lint"
@salesforce/soql-common: yarn run v1.22.10
@salesforce/soql-common: $ eslint ./src
@salesforce/soql-common: Done in 3.29s.
@salesforce/soql-model: yarn run v1.22.10
@salesforce/soql-model: $ eslint src/
@salesforce/soql-model: /Users/rbrainard/Development/soql-tooling/packages/soql-model/src/model/impl/headerCommentsImpl.test.ts
@salesforce/soql-model:   24:46  error  Replace `(null·as·unknown)` with `null·as·unknown`            prettier/prettier
@salesforce/soql-model:   27:42  error  Replace `(undefined·as·unknown)` with `undefined·as·unknown`  prettier/prettier
@salesforce/soql-model: /Users/rbrainard/Development/soql-tooling/packages/soql-model/src/validators/dateValidator.ts
@salesforce/soql-model:   68:35  error  Insert `⏎···`  prettier/prettier
@salesforce/soql-model: ✖ 3 problems (3 errors, 0 warnings)
@salesforce/soql-model:   3 errors and 0 warnings potentially fixable with the `--fix` option.
@salesforce/soql-model: error Command failed with exit code 1.
@salesforce/soql-model: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run lint exited 1 in '@salesforce/soql-model'
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @ lint: `lerna run lint`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @ lint script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
[...]

PR #253, opened 2021-05-10
feat: update select all label without star

ryanbrainard:
### What does this PR do?

As discussed in https://github.com/forcedotcom/soql-tooling/pull/251#discussion_r629080337, the `*` look like a bullet point in the dropdown. This changes it to `ALL FIELDS` without the `*`.  Not sure if we want to take this direction or not, but offering this PR as a place for discussion. 

image

Note, when pushing this, Husky blocked me on a bunch of lint errors. I bypassed them here, but will push up another PR with a fix for those. 

### What issues does this PR fix or reference?

https://github.com/forcedotcom/soql-tooling/pull/251#discussion_r629080337

PR #255, opened 2021-05-10
chore: fix prettier issues

ryanbrainard:
### What does this PR do?

When I was trying to push https://github.com/forcedotcom/soql-tooling/pull/253, Husky blocked my push because `lerna run lint` was failing with a bunch of prettier issues. I'm wondering if I got this because the git hooks are new and not installed on other people's machines? 

To fix this, I ran `lerna run prettier`, which calls `prettier --write` in the packages. It did format the code in some ugly ways, but I think it's probably better to follow Prettier standards (or override them).

As an aside, should we also be running `lerna run lint` in CI to avoid this kind of problem?

Fixes https://github.com/forcedotcom/soql-tooling/issues/254


dehru:
@ryanbrainard - What's interesting is when i pull your branch and run `yarn lint` i get 23 errors.  When i run it on main i get 3 errors.  I'm wondering what's going on.

ryanbrainard:
@dehru Sorry for the delay here. I looked more into this and what's happening is that the 3 problems on `main` are errors, so `yarn run lint` is failing fast and exiting `1`. Once those errors were fix in this PR, it allows linting to continue and reports 23 warnings, but at least exits `0`. I'd recommend we pull this in and file a new issue for the remaining warnings. We might also want to tweak the reporting levels of these, as a type issue should probably be treated more severally than a formatting one.

dehru:
> @dehru Sorry for the delay here. I looked more into this and what's happening is that the 3 problems on `main` are fatal errors, so `yarn run lint` is failing fast and exiting `1`. Once those errors were fix in this PR, it allows linting to continue and reports 23 non-fatal warnings, but at least exits `0`. I'd recommend we pull this in and file a [new issue](https://github.com/forcedotcom/soql-tooling/issues/257) for the remaining warnings. We might also want to tweak the reporting levels of these, as a type issue should probably be treated more severally than a formatting one.

Ah yes, that makes sense.

dehru (approved):

